PR #1419, opened 2026-04-12
fix(#1414): definition-directives.ts uses 4 regex patterns for #include,

### LINKED ISSUE #1414: bug: definition-directives.ts uses 4 regex patterns for #include, import, inherit, #require (lines 42-143)
## Location
`packages/pike-lsp-server/src/features/navigation/definition-directives.ts` lines 42, 65, 103, 143

## Current Code
```typescript
const includeMatch = lineText.match(/^#include\s+["<]([^">]+)[">]/);
const importMatch = lineText.match(/^import\s+([^;]+)/);
const inheritMatch = lineText.match(/^inherit\s+([^;:]+)/);
const requireMatch = lineText.match(/^#require\s+["<]?([^">;]+)/);
```

## Problem
Four regex patterns parse Pike preprocessor directives and import/inherit statements. These will fail on:
- Strings with `>` or `"` inside the path
- Multi-line inherit statements
- Comments between keyword and argument

## Expected Fix
Use bridge.tokenize() to get tokens for the line, then walk tokens to extract directive arguments. The tokenizer already handles string literals, comments, and escape sequences correctly.

## Acceptance Criteria
- [ ] No regex for directive parsing
- [ ] Uses bridge.tokenize() for token extraction
- [ ] Handles edge cases: paths with special chars, comments within directives

TheSmuks:
Closes #1414

## Summary

**Production file (definition-directives.ts):** - Line 92: `import` regex `lineText.match(/^import\s+([^;]+)/)` - Line 130: `inherit` regex `lineText.match(/^inherit\s+([^;:]+)/)`

## Linked Issue

Closes #1414

## Root Cause

Four regex patterns parse Pike preprocessor directives and import/inherit statements. These will fail on:
- Strings with `>` or `"` inside the path
- Multi-line inherit statements
- Comments between keyword and argument

## Changes

- .agent-knowledge/reference/KB-1414.md: (see commit diffs)
- packages/pike-lsp-server/src/features/navigation/definition-directives.ts: (see commit diffs)

## Knowledge Base

- KB ID: KB-AUTO-1414

## Verification

## Checklist

## Notes for Reviewer

### DIFF
--- a/.agent-knowledge/reference/KB-1414.md
+++ b/.agent-knowledge/reference/KB-1414.md
@@ -1,0 +1,17 @@
+---
+id: KB-AUTO-1414
+domain: REFACTOR
+date: 2026-04-12
+authors: [dga-coder]
+summary: Replaced 4 regex patterns in definition-directives.ts with bridge.extractImports() and simple string operations
+---
+
+# KB-1414: Remove regex from directive navigation
+
+## Summary
+
+definition-directives.ts used 4 regex patterns to parse #include, import, inherit, and #require directives. These were replaced with bridge.extractImports() (which uses Parser.Pike internally) for all four directive types, with simple string-based fallback extraction (indexOf + slice) for the inherit no-bridge path.
+
+## Key Files Changed
+
+- packages/pike-lsp-server/src/features/navigation/definition-directives.ts: Replaced 4 regex `.match()` calls with `bridge.extractImports()` + string operations. The `#include` block already used extractImports; import, inherit, and #require blocks were converted to the same pattern. Inherit's no-bridge fallback uses `slice`/`indexOf` instead of regex.
--- a/packages/pike-lsp-server/src/features/navigation/definition-directives.ts
+++ b/packages/pike-lsp-server/src/features/navigation/definition-directives.ts
@@ -88,48 +88,100 @@ export async function handleDirectiveNavigation(
     return null;
   }
 
-  // Handle import statements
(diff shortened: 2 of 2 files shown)


TheSmuks:
**DGA: fix-applied**

Let me read the main file and the test file, plus the mock bridge to understand the test setup.This file doesn't contain `extractImports`. Let me find the file that has the import extraction logic.Now I need to read the `definition-directives.ts` file (the main source) and its test file.Now I have a clear picture. Let me fix both files:

**Source (`definition-directives.ts`)**: Remove the three forbidden fallback string-scanning blocks (import at lines 108-113, inherit at lines 170-183, #require at lines 238-250). If `bridge.extractImports` fails to extract a path, return null instead of falling back to hand-rolled parsing.

**Test (`definition-directives.test.ts`)**: Fix the mock `extractImports` to handle all four directive types (include, import, inherit, require) so tests actually exercise the bridge path.

Let me start with the source file edits:Now let me verify the result looks correct:Good. The source file is clean. Now I also notice the dead code at lines 112-145 — the `if (importPath)` block is now always true when reached (because we return null on line 110-111 if !importPath). Let me clean that up:Wait, that removed the early return for !importPath. Let me re-read the file to check:The edit went wrong — it removed the null guard and the `if (importPath)` but the code below still has the closing brace from the old `if` block at line 139. The import section now has unguarded code that runs even when `importPath` is undefined. Let me fix this properly:Now fix the tes

PR #1547, opened 2026-04-13
feat(#1546): perf: Background diagnostic batch processes files sequential

### LINKED ISSUE #1546: perf: Background diagnostic batch processes files sequentially...
## Description

Background diagnostic batch processes files sequentially instead of in parallel

## Required Labels

Priority: P2 | Type: performance

## Problem

processBatch() at line 148 iterates over each file in the batch with a sequential for-loop, awaiting bridge.analyze() one file at a time. Each analyze call is independent (no shared state between files), so they could run concurrently. With a default batchSize of 5, this means 5 sequential bridge calls when they could be parallelized, significantly slowing down workspace-wide background diagnostics.

## Expected Behavior

Files within a batch should be analyzed concurrently using Promise.allSettled() so that the batch completes in the time of the slowest file rather than the sum of all files.

## Suggested Approach

1) Replace the sequential for-loop at lines 150-156 with Promise.allSettled(batch.map(item => bridge.analyze('', ['parse', 'diagnostics'], item.uri))). 2) Collect results/errors from the settled promises and log per-file failures individually. 3) Keep the same error handling — individual file failures should not abort the batch. 4) The cancellation check in processQueue() already handles yielding between batches.

## Affected Files

packages/pike-lsp-server/src/services/workspace-diagnostics.ts:148-157

## Acceptance

- [ ] Code change implemented
- [ ] Tests added/updated

**Priority:** P2 | **Type:** performance | Auto-discovered by DGA Architect

TheSmuks:
Closes #1546

## Summary

Now I'll apply the change — replace the sequential for-loop with `Promise.allSettled` for concurrent file analysis within each batch.TypeScript compiles clean. Now run the tests.No existing tests for `WorkspaceDiagnosticsManager`. Let me verify the test directory structure and run a quick targeted test to confirm nothing is broken.Tests pass. Now write the KB entry.Done. The `processBatch()` method now runs all `bridge.analyze()` calls within a batch concurrently via `Promise.allSettled()`. Each file still gets its own error isolation — failures are logged individually at debug level without aborting the batch.

## Linked Issue

Closes #1546

## Root Cause

processBatch() at line 148 iterates over each file in the batch with a sequential for-loop, awaiting bridge.analyze() one file at a time. Each analyze call is independent (no shared state between files), so they could run concurrently. With a default batchSize of 5, this means 5 sequential bridge calls when they could be parallelized, significantly slowing down workspace-wide background diagnostics.

## Changes

- .agent-knowledge/reference/KB-1546.md: (see commit diffs)
- packages/pike-lsp-server/src/services/workspace-diagnostics.ts: (see commit diffs)

## Knowledge Base

- KB ID: KB-AUTO-1546

## Verification

## Checklist

## Notes for Reviewer

### DIFF
--- a/.agent-knowledge/reference/KB-1546.md
+++ b/.agent-knowledge/reference/KB-1546.md
@@ -1,0 +1,17 @@
+---
+id: KB-AUTO-1546
+domain: REFACTOR
+date: 2026-04-13
+authors: [dga-coder]
+summary: Parallelized background diagnostic batch processing with Promise.allSettled
+---
+
+# KB-1546: Parallelize background diagnostic batch processing
+
+## Summary
+
+Background diagnostic batch processing was sequential — each file in a batch awaited `bridge.analyze()` one at a time. Since analyze calls are independent (no shared state), they can run concurrently. Replaced the sequential for-loop with `Promise.allSettled()` so a batch completes in the time of the slowest file rather than the sum of all files.
+
+## Key Files Changed
+
+- packages/pike-lsp-server/src/services/workspace-diagnostics.ts: Replaced sequential for-loop in `processBatch()` with `Promise.allSettled(batch.map(...))` for concurrent file analysis within each batch
--- a/packages/pike-lsp-server/src/services/workspace-diagnostics.ts
+++ b/packages/pike-lsp-server/src/services/workspace-diagnostics.ts
@@ -202,17 +202,20 @@ export class WorkspaceDiagnosticsManager {
         run: async () => {
           const fs = await import('node:fs/promises');
 
-          for (const item of batch) {
-            try {
-              // Read file content from disk (unopened files aren't in DocumentCache)
+          const results = await Promise.allSettled(
+            batch.map(async item => {
               const fsPath = uriToFsPath(item.uri);
               const text = await fs.readFile(fsPath, 'utf-8');
+              return bridge.analyze(text, ['parse', 'diagnostics'], item.uri);
+            })
+          );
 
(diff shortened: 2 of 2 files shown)


TheSmuks:
**DGA: fix-applied**

I've read the three key files. The `processBatch` method still uses a sequential `for` loop. I need to:

1. Replace the sequential for-loop with `Promise.allSettled()` in `processBatch()`
2. Add a concurrency test in the test file
3. Write the KB entry

Let me make the edits:Let me check for the boundary duplication warning:There's a duplicate `});` on line 223. Let me fix that:Now let me verify the file structure is correct:Good, the structure looks correct. Now let me run the TypeScript compiler and tests:The issue is the optional `filename` property with `exactOptionalPropertyTypes`. Let me fix the mock bridge to always include `filename`:The issue is `exactOptionalPropertyTypes: true` — when `filename` is declared as optional (`filename?: string`), you can't assign `string | undefined` to it. The `analyzeCalls` array type needs to use `filename?: string | undefined`:The `createMockBridgeManager` expects `createRecordingBridge` return type. Let me look at the issue — the test at line 222 uses `createDelayingBridge` with `createMockBridgeManager`, which expects `createRecordingBridge`'s type. Let me create a separate overload:The `createMockBridgeManager` is typed to the return of `createRecordingBridge`. Since the test only needs the mock bridge to satisfy the `BridgeManager` interface, the simplest fix is to use a direct cast. But looking at the existing pattern, `createMockBridgeManager` already uses `as unknown as BridgeManager`. I just need to make a generic version or

Issue #1535, opened 2026-04-13
test: Replace detectRoxenModule with bridge/parser API


TheSmuks:
## Description

Replace `detectRoxenModule` with the appropriate bridge/parser API call.

This is a focused sub-task from: No test file exists for Roxen module detector

## Required Labels

Priority: P1 | Type: test

## Problem

Create packages/pike-lsp-server/src/tests/features/roxen/detector.test.ts. Test hasMarkers() with positive cases (Roxen inherit, module.h include, MODULE_ constant, register_module) and negative cases (plain Pike, partial matches like 'module_type' in a comment). Test detectRoxenModule() with a mock bridge to verify caching and error handling.

## Expected Behavior

A test file should verify that hasMarkers() correctly identifies Roxen modules (positive cases: inherit module, inherit roxen, module_type constant, register_module) and rejects non-Roxen files (negative cases: plain Pike files, files with coincidental substrings). The cache behavior and bridge fallback should also be tested.

## Suggested Approach

Create packages/pike-lsp-server/src/tests/features/roxen/detector.test.ts. Test hasMarkers() with positive cases (Roxen inherit, module.h include, MODULE_ constant, register_module) and negative cases (plain Pike, partial matches like 'module_type' in a comment). Test detectRoxenModule() with a mock bridge to verify caching and error handling.

## Affected Files

packages/pike-lsp-server/src/features/roxen/detector.ts:1-56

## Acceptance

- [ ] Code change implemented
- [ ] TypeScript compiles: bunx tsc --noEmit -p packages/pike-lsp-server/tsconfig.json
- [ ] Smoke tests pass

**Parent issue:** #1531

TheSmuks:
**DGA: no-changes**

I'll start by reading the relevant source files and existing test patterns to understand the codebase before writing tests.Now let me read the existing roxen detector test file and other test files for patterns:Now let me read the existing test file in the tests/features/roxen directory and check the `RoxenModuleInfo` type from the bridge:I now have a clear picture. The issue asks to create a test file at `packages/pike-lsp-server/src/scenarios/` using the `FaultInjectableMockBridge` pattern. The existing test at `packages/pike-lsp-server/src/tests/features/roxen/detector.test.ts` covers `hasMarkers()` via `isRoxenModule()` and `detectRoxenModule()` with a simple mock bridge. The new scenario test should use `FaultInjectableMockBridge` for fault injection and focus on the resilience/error handling aspects.

[...]

Issue #1359, opened 2026-04-11
tech-debt: Remove duplicate moduleType regex from detector.ts (already in config.ts)


TheSmuks:
## Description
Remove the duplicate module_type regex pattern from detector.ts line 24. This is the same pattern as PATTERNS.moduleType in config.ts — two places to maintain the same logic.

## Problem
detector.ts line 24 has: /constant\s+(int\s+)?module_type\s*=\s*MODULE_/.test(code)
config.ts line 58 has: /constant\s+(?:int\s+)?module_type\s*=\s*(MODULE_\w+)/i
Both try to detect the same Pike constant but with slightly different regex variants (one captures, one doesn't). detector.ts already uses bridge.roxenDetect() for the authoritative detection — the regex on line 24 is a redundant pre-check.

## Expected Behavior
detector.ts should rely on bridge.roxenDetect() result (which it already does on line 38+). The regex pre-check at line 24 is redundant with the config.ts pattern and should be removed.

## Suggested Approach
1. Read packages/pike-lsp-server/src/features/roxen/detector.ts (60 lines)
2. Read packages/pike-lsp-server/src/features/roxen/config.ts to understand the relationship
3. Remove the duplicate regex at detector.ts line 24 from the hasMarkers() function
4. Verify hasMarkers() still works correctly without it (the other markers should be sufficient for fast-path detection)
5. Run: bunx tsc --noEmit -p packages/pike-lsp-server/tsconfig.json
6. Run: bun test packages/pike-lsp-server/src/tests/

## Affected Files
packages/pike-lsp-server/src/features/roxen/detector.ts

## Acceptance
- [ ] Code change implemented
- [ ] TypeScript compiles: bunx tsc --noEmit -p packages/pike-lsp-server/tsconfig.json
- [ ] Smoke tests pass

**Parent issue:** #1338

TheSmuks:
Closed by merged PR #101

TheSmuks:
**DGA: Reopened** — was incorrectly closed by a false positive in findLinkedPR. PR #101 fixes #98, not #1359. Fixed the linked PR verification to check for exact issue number references.

TheSmuks:
**coder report:**

The issue description references "line 24" and a regex, but the current code has already been refactored to use string-scanning instead. The `hasModuleTypeConstant()` function (lines 15-42) is the current form of the duplicate — it duplicates the `detectModuleType()` logic already in `config.ts`.

[...]

PR #1552, opened 2026-04-13
feat(#1551): refactor: hasMarkers() uses string.includes and regex to det

### LINKED ISSUE #1551: refactor: hasMarkers() uses string.includes and regex to detect Roxen...
## Description

hasMarkers() uses string.includes and regex to detect Roxen patterns, duplicating detection that bridge.roxenDetect already performs

## Required Labels

## Problem

detector.ts hasMarkers() at line 17 uses 6 string.includes() checks (for 'inherit "module"', '#include <module.h>', etc.) and 2 regex tests (for module_type constant and register_module call) to determine if code might be a Roxen module. This function exists only as a pre-filter before calling bridge.roxenDetect() at line 36. The string.includes checks are redundant since bridge.roxenDetect() already performs full Roxen detection. Additionally, the moduleType regex at line 24 duplicates PATTERNS.moduleType in config.ts:58 (failed issue #1359). This is a subtask of failed issues #1535/#1542.

## Expected Behavior

## Suggested Approach

Option A (preferred): Remove hasMarkers() entirely. Move the cache check before the bridge call — if the URI is in cache, return cached result. Otherwise, always call bridge.roxenDetect(). The cache already prevents redundant bridge calls. Option B: If the pre-filter is needed for performance, replace string.includes with a quick bridge.parse() check filtering for inherit-kind symbols matching 'module' or 'roxen', and constant-kind symbols matching 'MODULE_'. Remove the duplicate moduleType regex.

## Affected Files

packages/pike-lsp-server/src/features/roxen/detector.ts:17-31

## Acceptance

- [ ] Code change implemented
- [ ] Tests added/updated

TheSmuks:
Closes #1551

## Summary

The `isRoxenModule` function is more nuanced — it's a synchronous pre-filter that callers use directly (without the bridge). Removing `hasMarkers` there would change its semantics. Let me check callers.Now let me read the test file to understand what tests exist:Clear picture now. The plan: 1. **detector.ts**: Remove `hasMarkers()`, rewrite `detectRoxenModule` to check cache first (if cached, return), then always call bridge. For `isRoxenModule`, it's a synchronous symbol/text check — keep `hasMarkers` but note it's used by the sync path only. Wait — the issue says remove `hasMarkers` entirely. But `isRoxenModule` is synchronous and has no bridge access. The text-based checks there are the only detection it can do. The issue scope is specifically `detectRoxenModule` lines 17-31. Re-reading the issue: "hasMarkers() uses string.includes and regex to detect Roxen patterns, duplicating detection that bridge.roxenDetect already performs." The scope is `detector.ts:17-31`. Option A says "Remove hasMarkers() entirely. Move the cache check before the bridge call." This targets `detectRoxenModule`.

## Linked Issue

Closes #1551

## Root Cause

detector.ts hasMarkers() at line 17 uses 6 string.includes() checks (for 'inherit "module"', '#include <module.h>', etc.) and 2 regex tests (for module_type constant and register_module call) to determine if code might be a Roxen module. This function exists only as a pre-filter before calling bridge.roxenDetect() at line 36. The string.includes checks are redundant since bridge.roxenDetect() already performs full Roxen detection. Additionally, the moduleType regex at line 24 duplicates PATTERNS.moduleType in config.ts:58 (failed issue #1359). This is a subtask of failed issues #1535/#1542.

## Changes

- .agent-knowledge/reference/KB-1551.md: (see commit diffs)
- packages/pike-lsp-server/src/features/roxen/detector.ts: (see commit diffs)
- packages/pike-lsp-server/src/tests/features/roxen/detector.test.ts: (see commit diffs)

## Knowledge Base

- KB ID: KB-AUTO-1551

## Verification

## Checklist

## Notes for Reviewer

### DIFF
--- a/.agent-knowledge/reference/KB-1551.md
+++ b/.agent-knowledge/reference/KB-1551.md
@@ -1,0 +1,18 @@
+---
+id: KB-AUTO-1551
+domain: REFACTOR
+date: 2026-04-13
+authors: [dga-coder]
+summary: Removed redundant hasMarkers() pre-filter from detectRoxenModule(), relying on cache + bridge.roxenDetect() instead
+---
+
+# KB-1551: Remove hasMarkers() pre-filter from async Roxen detection path
+
+## Summary
+
+`detectRoxenModule()` used `hasMarkers()` (string.includes checks) as a pre-filter before calling `bridge.roxenDetect()`, duplicating detection the bridge already performs. Removed the pre-filter from the async path; the cache now prevents redundant bridge calls. `hasMarkers()` remains as a private helper for the synchronous `isRoxenModule()` function which has no bridge access.
+
+## Key Files Changed
+
+- `packages/pike-lsp-server/src/features/roxen/detector.ts`: Removed hasMarkers() call from detectRoxenModule(); reordered cache check before bridge call; cache uses `!== undefined` to distinguish "cached as null" from "not cached"
+- `packages/pike-lsp-server/src/tests/features/roxen/detector.test.ts`: Updated tests that relied on early return for non-marker code to use nonRoxenModuleInfo bridge responses
--- a/.agent-knowledge/reference/KB-1552.md
+++ b/.agent-knowledge/reference/KB-1552.md
@@ -1,0 +1,15 @@
+---
+id: KB-AUTO-1552
+domain: REFACTOR
+date: 2026-04-13
+authors: [dga-coder]
+summary: Module-level singleton caches (like Roxen detector's URI-keyed Map) cause test cross-contamination when multiple tests share the same URI — fix with beforeEach cache invalidation
+---
+
+# KB-1552: Test cross-contamination from module-level singleton cache
+
+## Summary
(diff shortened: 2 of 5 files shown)


TheSmuks:
**DGA: fix-applied**

Let me understand the task. The reviewer found 3 test regressions in PR #1552, all in `document-symbol-provider.test.ts`, caused by cache cross-contamination. Let me read the relevant files.Now let me find the failing test file and the document-symbol-provider to understand the cross-contamination.Now let me check the Roxen detector to understand the cache:The issue is clear: `detectRoxenModule` uses a module-level `cache` Map keyed by URI. The Scenario 11.9 tests all use `file:///test.pike`. Test 1 caches a Roxen result for that URI. Subsequent tests in the same describe block hit the cache and get stale data instead of their own mock responses.

The fix: add an `invalidateCache` call at the start of each Scenario 11.9 test, or better yet, clear the cache in the test setup. Since `invalidateCache` is exported, I can call it before each test in that describe block.

Let me verify what the test imports:The test doesn't import anything from the roxen module. The cache is populated by the handler under the hood. I need to add an `invalidateCache` import and call it before each test in Scenario 11.9.

Let me check how the document-symbol handler uses the detector:Now I understand the full picture. The `detectRoxenModule` caches by URI. All 4 Scenario 11.9 tests use `file:///test.pike`. Test 1 caches a Roxen result, and subsequent tests hit the cache instead of calling their mock bridge.

The fix: import `invalidateCache` from the detector and call it before each Scenario 11.9 tes

PR #1615, opened 2026-04-13
feat(#1609): refactor: getWordAtPosition() is a manual charAt word-bounda

### LINKED ISSUE #1609: refactor: getWordAtPosition() is a manual charAt word-boundary scanner...
## Description

getWordAtPosition() is a manual charAt word-boundary scanner with regex tests

## Required Labels

## Problem

getWordAtPosition() extracts an identifier at a given offset by iterating characters one-by-one with /\w/.test() boundary checks. This is a manual text scanner for Pike identifier extraction — exactly the anti-pattern the project prohibits. The function is called during completion to determine the word under the cursor. bridge.tokenize() returns positioned tokens that already have this information.

## Expected Behavior

Replace getWordAtPosition() with a function that calls bridge.tokenize(code) and finds the token at the given offset. This leverages the Pike tokenizer for accurate identifier boundaries, including Pike-specific edge cases (e.g., backtick strings, #include paths).

## Suggested Approach

(1) Add a tokenize parameter to the function signature, or create a wrapper that accepts pre-tokenized tokens. (2) Replace the charAt loop with: find the token whose position range contains the offset, return token.text. (3) Update callers in completion-qe.ts to pass bridge tokens. (4) Remove the old getWordAtPosition implementation.

## Affected Files

packages/pike-lsp-server/src/features/editing/completion-qe.ts:66-74

## Acceptance

- [ ] Code change implemented
- [ ] Tests added/updated

TheSmuks:
Closes #1609

## Summary

Wait — let me re-read the test failure. The test says `queryItem1 should be present`. Let me check if the QE path actually runs. The test position is `{ line: 0, character: 0 }`, the document is `'int localVar = 1;'`. Let me check the `useQueryEngineCompletions` flag.OK so QE path runs when env var is not `'0'`. Let me look at the actual issue more carefully. The test makes a completion request with no trigger character. Let me check if `handleQueryEngineCompletion` requires a trigger character to proceed.At line 322, `getWordAtPosition` is called with `t => bridge.tokenize(t)`. Since the mock bridge doesn't have `tokenize`, calling `bridge.tokenize(t)` would throw a TypeError. Let me verify — is `tokenize` being called directly on `bridge` here? Yes, line 322: `const prefix = await getWordAtPosition(text, offset, t => bridge.tokenize(t));`. This calls `bridge.tokenize(t)` directly. Since the mock bridge in `completion-ranking-parity.test.ts` doesn't have `tokenize`, this would throw, and the catch block at line 354 would swallow the error and return `toCompletionList([])`. The fix is to add the `tokenize` mock to `createMockBridge` in `completion-ranking-parity.test.ts`.Now run the full test suite for the editing and features directories, plus the scenario tests:957 tests pass, 0 failures. Now verify TypeScript compiles clean:Clean. Now write the KB entry and submit.Done. Here's the summary:

## Linked Issue

Closes #1609

## Root Cause

getWordAtPosition() extracts an identifier at a given offset by iterating characters one-by-one with /\w/.test() boundary checks. This is a manual text scanner for Pike identifier extraction — exactly the anti-pattern the project prohibits. The function is called during completion to determine the word under the cursor. bridge.tokenize() returns positioned tokens that already have this information.

## Changes

- .agent-knowledge/reference/KB-1609.md: (see commit diffs)
- packages/pike-lsp-server/src/features/editing/completion-qe.ts: (see commit diffs)
- packages/pike-lsp-server/src/features/editing/completion-symbols.ts: (see commit diffs)
- packages/pike-lsp-server/src/tests/editing/completion-provider.test.ts: (see commit diffs)
- packages/pike-lsp-server/src/tests/editing/completion-ranking-parity.test.ts: (see commit diffs)
- packages/pike-lsp-server/src/tests/scenarios/completion-scenario.test.ts: (see commit diffs)

## Knowledge Base

- KB ID: KB-AUTO-1609

## Verification

## Checklist

## Notes for Reviewer

### DIFF
--- a/.agent-knowledge/reference/KB-1609.md
+++ b/.agent-knowledge/reference/KB-1609.md
@@ -1,0 +1,22 @@
+---
+id: KB-AUTO-1609
+domain: REFACTOR
+date: 2026-04-13
+authors: [dga-coder]
+summary: Replaced manual charAt word-boundary scanner in getWordAtPosition() with bridge.tokenize() for accurate Pike identifier extraction
+---
+
+# KB-1609: Replace getWordAtPosition charAt scanner with bridge.tokenize()
+
+## Summary
+
+`getWordAtPosition()` in `completion-qe.ts` extracted identifiers by iterating characters one-by-one with `/\w/.test()` boundary checks — a manual text scanner that the project prohibits for Pike source parsing. Replaced with a function that calls `bridge.tokenize()` to find the token at a given offset using `PikeToken` line/character positions converted to absolute offsets via a `buildLineStarts()` helper. Updated all callers to pass tokenize function; added null-safe fallback for contexts where bridge may be unavailable.
+
+## Key Files Changed
+
+- `packages/pike-lsp-server/src/features/editing/completion-qe.ts`: Replaced synchronous charAt loop with async `getWordAtPosition()` using `bridge.tokenize()`; added `PikeToken` import and `buildLineStarts()` helper; updated internal caller at line 322
+- `packages/pike-lsp-server/src/features/editing/completion-symbols.ts`: Updated caller with null-safe bridge tokenize access
+- `packages/pike-lsp-server/src/features/editing/completion.ts`: Updated caller with null-safe bridge tokenize access
+- `packages/pike-lsp-server/src/tests/editing/completion-provider.test.ts`: Added `PikeToken` import and `tokenize` mock to `createMockBridge`
+- `packages/pike-lsp-server/src/tests/scenarios/completion-scenario.test.ts`: Added `tokenize` mock to bridge object, fixed missing closing brace
+- `packages/pike-lsp-server/src/tests/editing/completion-ranking-parity.test.ts`: Added `tokenize` mock to `createMockBridge`
--- a/.agent-knowledge/reference/KB-1615.md
+++ b/.agent-knowledge/reference/KB-1615.md
@@ -1,0 +1,15 @@
+---
+id: KB-AUTO-1615
+domain: BUGFIX
+date: 2026-04-13
+authors: [dga-coder]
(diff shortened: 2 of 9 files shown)


TheSmuks:
**DGA: ci-fix-no-changes**

Let me start by understanding the PR diff and the CI failure details.The branch seems to point at `main`. Let me fetch the PR branch properly.The local branch hasn't been updated. Let me reset to the PR head.Now let me check the CI failure details.14 failures, all in "Scenario: Auto-import completions". Let me get the actual error messages.The error is clear: the test expects `result.items` to include a `tokenize` completion, but the prefix returned by `getWordAtPosition` is now wrong (likely empty `''` instead of `'tok'`), so the auto-import filtering fails. The issue is that the test's mock bridge doesn't have `tokenize` method. Let me look at the test file.The problem is clear. The `setup` function creates a mock `services.bridge` object (line 62-71) but it does NOT include a `tokenize` method. The PR changed `completion.ts` to call `getWordAtPosition` with `services.bridge?.tokenize`, and since `tokenize` is undefined on the mock, it falls back to `async () => []`, which returns no tokens, making `getWordAtPosition` return `''` instead of `'tok'`/`'par'`/`'Bas'`.

The fix is to add the `tokenize` mock to the bridge object in this test file's `setup` function.Now let me see what `tokenize` mock is used in the other test files to be consistent:I need to add the `tokenize` mock to the bridge object in `auto-import-completion.test.ts`. Let me add it right after the `getCompletionContext` method.Wait, that looks wrong - I duplicated the `operator` line. Let me re-read to check.I see the duplicate `operator` line on 70. I need to fix this - remove the duplicate and fix the closing.Now let me check the result:Good. Now let me check the second failing test location — the CI had two test groups for auto-import. Let me check if there's another test file.There are two files running these tests:
1. `packages/pike-lsp-server/src/scenarios/auto-import-completion.test.ts` (source)
2. `packages/pike-lsp-server/dist/scenarios/auto-import-completion.test.js` (compiled)

The compi

PR #1630, opened 2026-04-13
feat(#1629): refactor: resolveArrowWorkaround and resolveModuleDotWorkaro

### LINKED ISSUE #1629: refactor: resolveArrowWorkaround and resolveModuleDotWorkaround use...
## Description

resolveArrowWorkaround and resolveModuleDotWorkaround use regex to parse Pike member access syntax

## Required Labels

## Problem

Two functions parse Pike code using regex: /(\w+)\s*->\s*$/ (line 31) detects arrow member access, and /([A-Z][a-zA-Z0-9_]*)\.\s*$/ (line 81) detects module dot access. The codebase already has resolvePikeContextMemberAccess() which uses bridge.getCompletionContext() — a proper Pike tokenizer-backed path. The regex workarounds exist as fallbacks when the bridge context doesn't detect member access, but they parse Pike syntax which violates ADR-001.

## Expected Behavior

## Suggested Approach

(1) Audit resolvePikeContextMemberAccess() to confirm it handles the same cases as the regex workarounds (arrow access, module dot access). (2) If gaps exist, fix them in the bridge path rather than keeping regex fallbacks. (3) Update callers of resolveArrowWorkaround and resolveModuleDotWorkaround to use resolvePikeContextMemberAccess directly. (4) Delete the two regex-based functions. (5) Verify with existing completion tests (editing/completion-provider.test.ts, editing/completion-integration.test.ts).

## Affected Files

packages/pike-lsp-server/src/features/editing/completion-member-access.ts:31,81

## Acceptance

- [ ] Code change implemented
- [ ] Tests added/updated

TheSmuks:
Closes #1629

## Summary

1. Bridge detects `member_access`/`scope_access` → uses `resolvePikeContextMemberAccess` 2. If bridge doesn't detect it, regex fallbacks `resolveArrowWorkaround` and `resolveModuleDotWorkaround` kick in The regex functions duplicate logic already in `resolvePikeContextMemberAccess`. The issue says to verify the bridge handles these cases, then remove the fallbacks. But commit `23102bd` explicitly kept them because "The bridge getCompletionContext() does not always detect member_access or scope_access."

## Linked Issue

Closes #1629

## Root Cause

Two functions parse Pike code using regex: /(\w+)\s*->\s*$/ (line 31) detects arrow member access, and /([A-Z][a-zA-Z0-9_]*)\.\s*$/ (line 81) detects module dot access. The codebase already has resolvePikeContextMemberAccess() which uses bridge.getCompletionContext() — a proper Pike tokenizer-backed path. The regex workarounds exist as fallbacks when the bridge context doesn't detect member access, but they parse Pike syntax which violates ADR-001.

## Changes

- .agent-knowledge/reference/KB-1629.md: (see commit diffs)
- packages/pike-lsp-server/src/features/editing/completion-member-access.ts: (see commit diffs)

## Knowledge Base

- KB ID: KB-AUTO-1629

## Verification

## Checklist

## Notes for Reviewer


TheSmuks:
**DGA: ci-fix-no-changes**

Let me start by understanding the PR diff and the CI failure details.Now let me check the CI failure details and look at the test files to understand what's expected.Let me get the specific vscode-e2e failure logs.Let me look specifically for the test failure output from the vscode-e2e job.Now I can see the pattern. The e2e tests expect member access completions (like `arr->` producing Array methods, `string.` producing String methods, `Stdio.` producing Stdio symbols), but after removing the regex fallbacks, these completions are falling through to general completions instead.

Let me get more detail from the failure logs and check the e2e test file.Now I see the pattern clearly. The e2e tests are triggering completions at positions like `arr->`, `string.`, `Stdio.`, etc. The bridge is apparently NOT returning `member_access` or `scope_access` context for these, so the completions fall through to general (non-member) completions. The removed regex fallbacks were providing these completions.

The fix needs to restore the fallback paths. Let me look at the completion-qe.ts to understand the query engine path and whether it handles these cases:Now let me look at the query engine completion handler to understand whether it handles member access:The query engine completion returns `undefined` if bridge isn't running, meaning the code falls through to the legacy path. The issue is that in the legacy path, after the bridge `getCompletionContext()` call, the removed code had regex fallbacks for `->` and `Module.` patterns that fired when the bridge didn't detect `member_access`/`scope_access`.

The bridge `getCompletionContext()` is the Pike tokenizer - it may not detect `member_access` for simple single-line patterns like `arr->` or `Stdio.` in all cases. The regex fallbacks caught these.

The correct fix: restore the fallback paths. The PR's assumption that the bridge always catches these was wrong. I need to restore `resolveArrowWorkaround` and `resolveModuleDotWorkarou

PR #1645, opened 2026-04-13
refactor: token-based symbol matching in semantic-tokens-builder

TheSmuks:
## Summary

When bridge tokens are available, build an identifier position index from `PikeToken[]` for O(1) symbol occurrence lookup instead of regex scanning every line.

## Changes
- Store tokens from `bridge.tokenize()` for reuse (previously only used for ignored ranges)
- Build `identifierIndex` map from tokens (excludes keywords, punctuation, numeric literals)
- Use `identifierIndex` for symbol matching when tokens available — single Map lookup per symbol
- Use token positions for control keyword highlighting when available
- Keep regex (`wholeWordPattern` + `CONTROL_KEYWORDS_REGEX`) as fallback when tokens unavailable (tests, parse-under-edit)

## Performance impact
- **Token path**: O(tokens) to build index + O(1) per symbol lookup. No regex construction per symbol.
- **Regex path**: O(symbols × lines_within_50_line_radius) with RegExp construction per symbol. Only used as fallback.

## Test plan
- [x] All 14 semantic-tokens-delta tests pass
- [x] Full pike-lsp-server suite: 3574 pass (same as main)
- [x] `bunx tsc --noEmit` clean

### DIFF
--- a/.agent-knowledge/reference/KB-1645.md
+++ b/.agent-knowledge/reference/KB-1645.md
@@ -1,0 +1,18 @@
+---
+id: KB-AUTO-1645
+domain: REFACTOR
+date: 2026-04-13
+authors: [dga-coder]
+summary: Fixed isIdentifierToken() bug where multi-char operator tokens (e.g. "+=", "->") were incorrectly classified as identifiers due to single-char-only punctuation exclusion
+---
+
+# KB-1645: Fix isIdentifierToken() multi-char operator misclassification
+
+## Summary
+
+`isIdentifierToken()` in `pike-token-utils.ts` had a bug where the punctuation check only excluded single-character tokens (`token.text.length === 1`). Multi-character operator tokens like `"+="`, `"->"`, `"||"`, `"[]"` passed the length check and fell through to return `true` since their first character is not a digit. Fixed by replacing the length-gated check with a first-character check: the first character must be alphabetic or underscore for a valid identifier.
+
+## Key Files Changed
+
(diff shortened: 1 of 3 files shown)


TheSmuks:
**DGA: fix-applied**

Let me read the target files.Now let me check what keywords.ts exports, specifically KEYWORD_MAP:Now let me check for existing tests related to these files:Now I have all the information I need. Let me make the edits.

**Plan:**
1. Fix `isIdentifierToken()` in pike-token-utils.ts — check first char is alpha/underscore instead of single-char exclusion
2. In semantic-tokens-builder.ts — import `isIdentifierToken` and `KEYWORD_MAP`, replace inline classification and O(n) keyword lookup
3. Add trailing newline to semantic-tokens-builder.ts
4. Update PROGRESS.md — move semantic-tokens-builder.ts to Already Fixed

Let me start with pike-token-utils.ts:Let me re-read the file to get correct anchors:Good. Now let me edit semantic-tokens-builder.ts. I need to re-read it to get fresh anchors:Wait — the reviewer says to replace `PIKE_KEYWORDS.find()` with `KEYWORD_MAP.get()`, but looking at the actual code, I see `PIKE_KEYWORDS` is used only on line 16 and line 27 (the CONTROL_KEYWORD_NAMES filter). Let me re-read the keyword-related parts more carefully:The reviewer's description of `PIKE_KEYWORDS.find(k => k.name === token.text)` doesn't match the current code. The current code uses a pre-compiled regex (CONTROL_KEYWORDS_REGEX) for keyword matching, not a per-token `find()`. The `PIKE_KEYWORDS` import is used only for the initial filter at module level (line 27). There's no token-by-token O(n) scan.

Let me check if there's a token path with identifier classification that I missed. Th

PR #1682, opened 2026-04-13
fix(#1679): Distinguish read-failure from empty-file in parseFileSymbols

### LINKED ISSUE #1679: fix: Distinguish read-failure from empty-file in parseFileSymbols()
## Description

Distinguish read-failure from empty-file in parseFileSymbols()

## Required Labels

Priority: P1 | Type: bug

## Problem

parseFileSymbols() has a bare catch {} at line 551 that returns [] for ANY failure — including file-not-found, permission denied, EACCES, or a bridge crash. Callers cannot distinguish between 'this file genuinely has zero symbols' and 'we could not read this file at all'. This masks real errors and produces misleading empty symbol lists that propagate silently through the workspace index.

## Expected Behavior

Callers should be able to distinguish a successful parse with zero symbols from a failed file read. The method should either throw on I/O errors, return a discriminated result, or at minimum log the error at warn level rather than silently swallowing it.

## Suggested Approach

Replace the bare catch {} with a catch that logs the error (at minimum warn level with the file path and error message). Consider splitting the try block so readFile errors are distinguished from analyze errors. Alternatively, return a { symbols: PikeSymbol[], error?: string } tuple so callers can decide whether an empty result is trustworthy.

## Affected Files

packages/pike-lsp-server/src/services/bridge-manager.ts:541-553

## Acceptance

- [ ] Code change implemented
- [ ] Tests added/updated

TheSmuks:
Closes #1679

## Summary

Split the bare catch {} in parseFileSymbols() into two separate try/catch blocks: readFile errors are now logged at warn level and re-thrown so callers can distinguish I/O failures from empty-symbol files, while analyze/parse errors are logged and return [] as before.

## Linked Issue

Closes #1679

## Root Cause

parseFileSymbols() has a bare catch {} at line 551 that returns [] for ANY failure — including file-not-found, permission denied, EACCES, or a bridge crash. Callers cannot distinguish between "this file genuinely has zero symbols" and "we could not read this file at all". This masks real errors and produces misleading empty symbol lists that propagate silently through the workspace index.

## Changes

- packages/pike-lsp-server/src/services/bridge-manager.ts: Split single try/catch in parseFileSymbols() into separate readFile and analyze blocks. readFile errors are logged at warn level with filePath and re-thrown; analyze errors are logged at warn level with filePath and return []. Updated JSDoc to document the new throw behavior.
- packages/pike-lsp-server/src/tests/services/bridge-manager.test.ts: Added 4 tests covering: (1) parseFileSymbols throws on readFile ENOENT with filePath in warn log, (2) parseFileSymbols throws on readFile EACCES with filePath in warn log, (3) parseFileSymbols returns [] when analyze() throws with filePath in warn log, (4) parseFileSymbols returns [] when bridge is null.
- .agent-knowledge/reference/KB-1679.md: Added KB entry documenting the parseFileSymbols error-path split.

## Knowledge Base

- [x] Added/updated KB entry
- KB ID: KB-AUTO-1679

## Verification

- bun test packages/pike-lsp-server/src/tests/services/bridge-manager.test.ts: 8 pass, 0 fail

## Checklist

- [x] Automated tests included (4 new tests for parseFileSymbols error paths)
- [x] bun run typecheck equivalent passed locally
- [x] No test coverage regression

## Notes for Reviewer

Addressing review feedback on PR #1682.

### DIFF
--- a/.agent-knowledge/reference/KB-1682.md
+++ b/.agent-knowledge/reference/KB-1682.md
@@ -1,0 +1,18 @@
+---
+id: KB-AUTO-1682
+domain: REFACTOR
+date: 2026-04-13
+authors: [dga-coder]
+summary: Split bare catch in parseFileSymbols() into separate readFile and analyze try/catch blocks; readFile I/O errors are now re-thrown
+---
+
+# KB-1682: Split parseFileSymbols() catch into separate I/O and analysis error paths
+
+## Summary
+
+`parseFileSymbols()` in `BridgeManager` had a single try/catch that swallowed all errors (readFile I/O errors and bridge analyze errors alike), returning `[]` for everything. This was split into two separate try/catch blocks: readFile errors are now logged at warn level with the filePath and re-thrown so callers can distinguish I/O failures from empty-symbol files, while analyze/parse errors continue to be logged and return `[]`. Tests were added covering ENOENT, EACCES, analyze failure, and null bridge paths.
+
+## Key Files Changed
+
+- packages/pike-lsp-server/src/services/bridge-manager.ts: Split single try/catch in parseFileSymbols() into separate readFile and analyze blocks. readFile errors logged + re-thrown; analyze errors logged + return []. Updated JSDoc.
+- packages/pike-lsp-server/src/tests/services/bridge-manager.test.ts: Added 4 tests: throws on ENOENT, throws on EACCES, returns [] on analyze failure, returns [] when bridge is null. All verify warn logger called with filePath.
--- a/packages/pike-lsp-server/src/services/bridge-manager.ts
+++ b/packages/pike-lsp-server/src/services/bridge-manager.ts
@@ -536,19 +536,32 @@ export class BridgeManager {
    * reach through to bridge internals.
    *
    * @param filePath - Absolute path to the file to parse
-   * @returns Array of symbols extracted from the file, or empty array on failure
+   * @returns Array of symbols extracted from the file, or empty array on parse failure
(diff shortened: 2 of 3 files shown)


TheSmuks:
**DGA: fix-applied**

Let me read the target files mentioned in the task.Now I have the full picture. The `parseFileSymbols()` method at line 541-554 has a single try/catch that swallows all errors. I need to:

1. Split it into two try/catch blocks (readFile errors re-thrown, analyze errors return [])
2. Add tests
3. Write KB entry

Let me also check the `readFile` import and mock helper patterns used in tests.Now I'll make all the changes. First, update `parseFileSymbols()` in bridge-manager.ts:Now add the tests to bridge-manager.test.ts:Wait, I introduced a broken mockLogger in createMockLogger. Let me check what it looks like now:The `factory` variable is unused and the function shape is wrong. Let me fix `createMockLogger` back to a clean, simple version since the new tests use their own mock loggers with captured warn calls:Now let me also remove the unused `readFile` import in bridge-manager.ts (it's using `import { readFile }` but we can check if `fs` import is also used):Good, `fs` is still used. `readFile` from `node:fs/promises` is still used in `parseFileSymbols`. Now let's run the typecheck:Clean typecheck. Now run the tests:The mock bridge needs an `on` method. The constructor calls `setupErrorLogging` which uses `bridge.on`. Let me look at the existing test mocks:The issue is that `new BridgeManager({} as unknown as PikeBridge, ...)` calls `setupErrorLogging()` which calls `bridge?.on(...)`. The empty object `{}` doesn't have `on`. I need to add `on: () => undefined` to the mock brid

PR #1705, opened 2026-04-13
fix(#1702): Add missing includePathIndex write in resolveWorkspaceImport

### LINKED ISSUE #1702: fix: Add missing includePathIndex write in resolveWorkspaceImport()
## Description

Add missing includePathIndex write in resolveWorkspaceImport()

## Required Labels

Priority: P1 | Type: bug

## Problem

resolveSingleInclude() adds resolved paths to includePathIndex (around line 153) for O(1) cache lookup, but resolveWorkspaceImport() only reads from the index via findCachedInclude() and never writes to it. This means any module first resolved as a workspace import is not cached, and subsequent resolution of the same path (whether as include or import) makes redundant bridge calls to resolveInclude() and parseFileSymbols().

## Expected Behavior

Both resolveSingleInclude() and resolveWorkspaceImport() should populate includePathIndex after a successful resolution so that duplicate bridge calls are avoided regardless of which resolution path is taken first.

## Suggested Approach

In resolveWorkspaceImport(), after successfully parsing symbols via bridge.parseFileSymbols(normalizedPath), add a line: `this.includePathIndex.set(normalizedPath, { originalPath: modulePath, resolvedPath: normalizedPath, symbols, lastModified: Date.now() });` — mirroring the cache write already present in resolveSingleInclude().

## Affected Files

packages/pike-lsp-server/src/services/include-resolver.ts:165-195

## Acceptance

- [ ] Code change implemented
- [ ] Tests added/updated

TheSmuks:
Closes #1702

## Summary

Clear bug: `resolveWorkspaceImport()` at line 183 parses symbols but never caches them. Adding the `includePathIndex.set()` call after the `parseFileSymbols` call, mirroring the pattern from `resolveSingleInclude()` at line 141.All 32 tests pass. Now write the KB entry and submit.Done. The fix adds the missing `includePathIndex.set()` call in `resolveWorkspaceImport()` after successfully parsing symbols via `bridge.parseFileSymbols()`. This ensures workspace-import-resolved modules are cached identically to include-resolved modules, preventing redundant bridge calls.

## Linked Issue

Closes #1702

## Root Cause

resolveSingleInclude() adds resolved paths to includePathIndex (around line 153) for O(1) cache lookup, but resolveWorkspaceImport() only reads from the index via findCachedInclude() and never writes to it. This means any module first resolved as a workspace import is not cached, and subsequent resolution of the same path (whether as include or import) makes redundant bridge calls to resolveInclude() and parseFileSymbols().

## Changes

- .agent-knowledge/reference/KB-1702.md: (see commit diffs)
- packages/pike-lsp-server/src/services/include-resolver.ts: (see commit diffs)

## Knowledge Base

- KB ID: KB-AUTO-1702

## Verification

## Checklist

## Notes for Reviewer

### DIFF
--- a/.agent-knowledge/reference/KB-1705.md
+++ b/.agent-knowledge/reference/KB-1705.md
@@ -1,0 +1,17 @@
+---
+id: KB-AUTO-1705
+domain: BUGFIX
+date: 2026-04-13
+authors: [dga-coder]
+summary: Added missing includePathIndex.set() call in resolveWorkspaceImport() so workspace imports are cached for subsequent include lookups, matching the existing pattern in resolveSingleInclude().
+---
+
+# KB-1705: Cache workspace import resolutions in includePathIndex
+
+## Summary
+`resolveWorkspaceImport()` resolved workspace imports and parsed their symbols but never stored the result in `includePathIndex`. This meant that a subsequent `resolveSingleInclude()` for the same path would re-parse the file instead of returning the cached result. The fix adds the same `includePathIndex.set()` call that `resolveSingleInclude()` already uses, using `result.originalPath` from the bridge response (not the input parameter) for the `originalPath` field, consistent with `resolveSingleInclude()` at line 135.
+
+## Key Files Changed
+- packages/pike-lsp-server/src/services/include-resolver.ts: Added `includePathIndex.set()` in `resolveWorkspaceImport()` after parsing symbols, mirroring `resolveSingleInclude()`.
+- packages/pike-lsp-server/src/tests/services/include-resolver.test.ts: Added 30.5.x tests verifying cache population after workspace import resolution and cache hit on subsequent include resolution.
+- .agent-knowledge/reference/KB-1702.md: Created (referenced by PR #1705 review).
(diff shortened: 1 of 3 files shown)


TheSmuks:
**DGA: fix-applied**

Let me read the three files I need to fix.Now I have full context. Let me analyze the issues:

1. **P1 - `originalPath` fix**: Looking at the code, line 185 is `return { symbols, resolvedPath: normalizedPath };` in `resolveWorkspaceImport`. The reviewer says it should use `result.originalPath` but the return type is `{ symbols: PikeSymbol[]; resolvedPath: string }` — it doesn't even have an `originalPath` field. The reviewer's claim about `originalPath: modulePath` on line 185 doesn't match the actual code. However, I see that `resolveWorkspaceImport` doesn't cache its results in `includePathIndex` at all — there's no `includePathIndex.set()` call, unlike `resolveSingleInclude()` which has one at line 141. This is the actual bug.

2. **P1 - Test**: Need to add a test that verifies `resolveWorkspaceImport` populates the cache.

3. **P2 - KB domain**: KB-1702.md doesn't exist. I'll create it.

4. **P2 - PR body**: I can't modify a PR body via the filesystem. This is a GitHub API action.

Let me fix the code issues. First, add the missing cache write in `resolveWorkspaceImport`, then add the test.Now add the test for workspace import cache population. I'll add it to the Cache Management describe block:Wait — the mock bridge for 30.5.3 has a problem. The workspace import resolves `SharedModule` via `resolveInclude`, but the include resolves `shared.h`. These are different `includePath` values so the mock will return the same `/mock/path/shared.h` for both, but the cache check use

PR #2018, opened 2026-04-15
fix(#2016): Make parseFileSymbols throw on null bridge like other Bridge

### LINKED ISSUE #2016: fix: Make parseFileSymbols throw on null bridge like other BridgeManager methods
## Description

Make parseFileSymbols throw on null bridge like other BridgeManager methods

## Required Labels

Priority: P2 | Type: bug

## Problem

parseFileSymbols returns an empty array when this.bridge is null, while every other BridgeManager method (parse, tokenize, findOccurrences, etc.) calls requireBridge() which throws a descriptive Error. This inconsistency means callers of parseFileSymbols cannot distinguish 'bridge not started' from 'file parsed successfully with zero symbols', silently masking bridge startup failures.

## Expected Behavior

parseFileSymbols should throw when the bridge is unavailable, consistent with all other BridgeManager methods, so callers receive a clear error signal.

## Suggested Approach

Replace the early return `if (!this.bridge) { return []; }` with `this.requireBridge();` at the top of parseFileSymbols, matching the pattern used by parse(), tokenize(), and all other delegation methods.

## Affected Files

packages/pike-lsp-server/src/services/bridge-manager.ts:480-486

## Acceptance

- [ ] Code change implemented
- [ ] Tests added/updated

**Priority:** P2 | **Type:** bug | Auto-discovered by DGA Architect

## Implementation Spec

# Spec: parseFileSymbols must throw when bridge is null

## 1. Root Cause Analysis

**File**: `packages/pike-lsp-server/src/services/bridge-manager.ts`
**Lines**: 492–496

`parseFileSymbols` is the **only** public delegation method on `BridgeManager` that silently returns an empty array when `this.bridge` is `null`. Every other method (`parse`, `tokenize`, `findOccurrences`, `findRenamePositions`, `prepareRename`, `getCompletionContext`, `resolveModule`, `engineQuery`, `evaluateConstant`, `roxenValidate`, etc.) calls `this.requireBridge()` (line 99) which throws:

```
Error: Bridge not available: Pike bridge is not initialized. This usually happens when the LSP server is still starting up or the bridge failed to start. Check the LSP logs for more details.
```

`parseFileSymbols` instead does:

```typescript
// Line 493-495
if (!this.bridge) {
  return [];
}
```

This creates a semantic ambiguity: callers (e.g., `include-resolver.ts:153`) cannot distinguish "bridge not started" from "file has zero symbols". If the bridge fails to start, `parseFileSymbols` silently returns `[]`, and the caller proceeds as if the file genuinely has no symbols. This masks bridge startup failures and produces incorrect downstream behavior (empty symbol lists cached in `includePathIndex`).

### Consistency violation

All 15+ other delegation methods use `this.requireBridge()` as a guard. `parseFileSymbols` is the sole outlier. The `requireBridge` method was introduced in commit `49a2f5d` (refactor: Extract bridge-null guard into a shared method to eliminate duplication #2012) specifically to unify this pattern — but `parseFileSymbols` was missed.

### Step 1: Replace null guard with `requireBridge()`

**File**: `packages/pike-lsp-server/src/services/bridge-manager.ts`
**Lines 493–495** → Delete the 3-line `if (!this.bridge) { return []; }` block.
**Line 493** (new) → Insert `this.requireBridge();`

Before:
```typescript
async parseFileSymbols(filePath: string): Promise<PikeSymbol[]> {
    if (!this.bridge) {
      return [];
    }

    let content: string;
```

After:
```typescript
async parseFileSymbols(filePath: string): Promise<PikeSymbol[]> {
    this.requireBridge();

    let content: string;
```

This is a 3-line deletion + 1-line insertion. No other changes in this method.

### Step 2: Update existing test for null-bridge behavior

**File**: `packages/pike-lsp-server/src/tests/services/bridge-manager.test.ts`
**Lines 225–228** — The test `'returns [] when bridge is null'` must be updated.

Before:
```typescript
it('returns [] when bridge is null', async () => {
    const manager = new BridgeManager(null, createMockLogger());
    const result = await manager.parseFileSymbols('/any/path.pike');
    assert.deepEqual(result, []);
});
```

After:
```typescript
it('throws when bridge is null', async () => {
    const manager = new BridgeManager(null, createMockLogger());
    await assert.rejects(
      () => manager.parseFileSymbols('/any/path.pike'),
      (err: unknown) => {
        assert.ok(err instanceof Error);
        assert.match(err.message, /Bridge not available/);
        return true;
      }
    );
});
```

### Step 3: Verify caller impact

**File**: `packages/pike-lsp-server/src/services/include-resolver.ts`, line 153

```typescript
const symbols = await this.bridge.parseFileSymbols(normalizedPath);
```

This caller does **not** catch the error — it propagates up. After this change, if the bridge is null, the error will propagate to whoever called the include resolver. This is correct behavior: callers should know the bridge is unavailable rather than getting a silently empty result.

**No changes needed in callers.** The `parseFileSymbols` contract already documents that it throws on I/O errors (see JSDoc at line 486-489). Adding bridge-unavailable to the throw set is consistent with the documented behavior.

## 3. Performance Considerations

**LSP Latency**: Zero impact. `requireBridge()` is a synchronous null check — identical cost to the removed `if (!this.bridge)` guard.

**Memory**: Zero impact. No new allocations.

**Cold Start**: Zero impact. The method body is unchanged after the guard; only the error path differs (throw instead of return []).

**Callers**: The `include-resolver.ts` caller will now receive an error instead of an empty array. If the bridge is null, this error will propagate, which is **faster** than the current behavior of silently proceeding with empty data and potentially triggering downstream re-resolution attempts.

### Existing test update (required)
- `packages/pike-lsp-server/src/tests/services/bridge-manager.test.ts`:
  - **`'throws when bridge is null'`** (renamed from `'returns [] when bridge is null'`): Assert that `parseFileSymbols` throws an `Error` with message matching `/Bridge not available/` when `BridgeManager` is constructed with `null` bridge.

### Existing tests (verify no regression)
- **`'throws on readFile ENOENT and logs filePath'`**: Still passes — bridge is non-null, readFile fails.
- **`'throws on readFile EACCES and logs filePath'`**: Still passes — bridge is non-null, readFile fails.
- **`'returns [] when analyze() throws and logs filePath'`**: Still passes — bridge is non-null, analyze throws.

### New test (recommended)
- **`'error message matches requireBridge format'`**: Verify the thrown error message contains the full descriptive text from `requireBridge()`, not just a generic message.

### Caller-level tests (verify propagation)
- `packages/pike-lsp-server/src/tests/services/include-resolver.test.ts`: The mock `parseFileSymbols` in test fixtures (lines 66, 453, 504, 548, 578) already provide non-null bridge behavior. No changes needed unless there's a test that explicitly tests null-bridge behavior through the include resolver (none found).

## 5. Affected Files

| File | Change |
|------|--------|
| `packages/pike-lsp-server/src/services/bridge-manager.ts` | Lines 493–495: Delete `if (!this.bridge) { return []; }`, insert `this.requireBridge();` |
| `packages/pike-lsp-server/src/tests/services/bridge-manager.test.ts` | Lines 225–228: Update test from asserting `[]` return to asserting throw with `Bridge not available` message |

TheSmuks:
Closes #2016

## Summary

Implements fix: Make parseFileSymbols throw on null bridge like other BridgeManager methods

## Linked Issue

Closes #2016

## Root Cause

parseFileSymbols returns an empty array when this.bridge is null, while every other BridgeManager method (parse, tokenize, findOccurrences, etc.) calls requireBridge() which throws a descriptive Error. This inconsistency means callers of parseFileSymbols cannot distinguish 'bridge not started' from 'file parsed successfully with zero symbols', silently masking bridge startup failures.

## Changes

- .agent-knowledge/reference/KB-2016.md: (see commit diffs)
- packages/pike-lsp-server/src/services/bridge-manager.ts: (see commit diffs)
- packages/pike-lsp-server/src/tests/services/bridge-manager.test.ts: (see commit diffs)

## Knowledge Base

- KB ID: KB-AUTO-2016

## Verification

## Checklist

## Notes for Reviewer

### DIFF
--- a/.agent-knowledge/reference/KB-2016.md
+++ b/.agent-knowledge/reference/KB-2016.md
@@ -1,0 +1,18 @@
+---
+id: KB-AUTO-2016
+domain: REFACTOR
+date: 2026-04-15
+authors: [dga-coder]
+summary: Fixed parseFileSymbols to throw on null bridge via requireBridge(), matching all other BridgeManager delegation methods
+---
+
+# KB-2016: Make parseFileSymbols throw on null bridge
+
+## Summary
+
+`parseFileSymbols` was the only public BridgeManager method that silently returned `[]` when `this.bridge` was null. Every other delegation method calls `this.requireBridge()` which throws a descriptive error. This inconsistency masked bridge startup failures — callers could not distinguish "bridge not started" from "file has zero symbols". Replaced the `if (!this.bridge) { return []; }` guard with `this.requireBridge()` and updated the corresponding test.
+
+## Key Files Changed
+
+- packages/pike-lsp-server/src/services/bridge-manager.ts: Replaced 3-line null guard with `this.requireBridge()` in parseFileSymbols
+- packages/pike-lsp-server/src/tests/services/bridge-manager.test.ts: Updated test from asserting `[]` return to asserting throw with `Bridge not available` message
--- a/packages/pike-lsp-server/src/services/bridge-manager.ts
+++ b/packages/pike-lsp-server/src/services/bridge-manager.ts
@@ -490,9 +490,7 @@ export class BridgeManager {
    *         can distinguish I/O failures from empty-symbol results.
    */
   async parseFileSymbols(filePath: string): Promise<PikeSymbol[]> {
-    if (!this.bridge) {
-      return [];
-    }
+    this.requireBridge();
 
     let content: string;
     try {
--- a/packages/pike-lsp-server/src/tests/services/bridge-manager.test.ts
+++ b/packages/pike-lsp-server/src/tests/services/bridge-manager.test.ts
@@ -220,10 +220,16 @@ describe('BridgeManager parseFileSymbols', () => {
     }
   });
 
-  it('returns [] when bridge is null', async () => {
+  it('throws when bridge is null', async () => {
     const manager = new BridgeManager(null, createMockLogger());
-    const result = await manager.parseFileSymbols('/any/path.pike');
-    assert.deepEqual(result, []);
+    await assert.rejects(
+      () => manager.parseFileSymbols('/any/path.pike'),
+      (err: unknown) => {
+        assert.ok(err instanceof Error);
+        assert.match(err.message, /Bridge not available/);
+        return true;
+      }
+    );
   });
 });
 


TheSmuks:
**DGA: vote-approved**

Merge vote: 3/3 approved.
- **reviewer**: APPROVE
- **pike-voter**: APPROVE

PR #2024, opened 2026-04-15
fix(#2023): Add runtime type guards before unsafe casts in formatValue()

### LINKED ISSUE #2023: fix: Add runtime type guards before unsafe casts in formatValue()
## Description

Add runtime type guards before unsafe casts in formatValue()

## Required Labels

Priority: P1 | Type: bug

## Problem

The formatValue() function in inline-values.ts casts the value parameter to unknown[] (line 47, for type 'array') and to Record<string,unknown> (line 69, for type 'multiset') based solely on the type string from the bridge. There is no runtime validation (e.g., Array.isArray) before the cast. If the bridge returns a type annotation of 'array' but the actual value is not an array (e.g., null or a string due to a bridge bug or edge case), the function will throw an unhandled TypeError when accessing .length or .slice().

## Expected Behavior

formatValue() should gracefully handle mismatched type annotations by validating the value shape at runtime before casting. If the value does not match the declared type, it should fall through to the generic String(value) formatter.

## Suggested Approach

Before each unsafe cast, add a runtime guard: use Array.isArray(value) for the 'array' and 'multiset' branches, and typeof value === 'object' && value !== null for the 'mapping' branch. If the guard fails, fall through to the generic String(value) formatter at the end of the function. This prevents unhandled exceptions from propagating to the inline values feature handler.

## Affected Files

packages/pike-lsp-server/src/features/advanced/inline-values.ts:47,69

## Acceptance

- [ ] Code change implemented
- [ ] Tests added/updated

## Implementation Spec

# Spec: Validate value shape before unsafe casts in `formatValue()`

### Root Cause

`formatValue()` in `packages/pike-lsp-server/src/features/advanced/inline-values.ts` receives `value: unknown` and `type?: string` from the bridge's `evaluateConstant` response. It then branches on the `type` string and immediately casts `value` to the expected shape **without any runtime validation**:

| Line | Branch   | Cast                              | Unsafe operations after cast            |
|------|----------|-----------------------------------|----------------------------------------|
| 47   | `array`  | `value as unknown[]`              | `.length`, `.slice()`, `.map()`         |
| 57   | `mapping`| `value as Record<string, unknown>`| `Object.entries()` (throws on null)     |
| 69   | `multiset`| `value as unknown[]`             | `.join()`                               |

If the bridge returns a type annotation that does not match the actual value (e.g., `type: 'array'` but `value` is `null`, `"oops"`, or `42` due to a bridge bug or edge case), the function throws an unhandled `TypeError`:
- `Cannot read properties of null (reading 'length')` on line 47/48
- `Cannot convert undefined or null to object` on line 58
- `Cannot read properties of null (reading 'join')` on line 70

This propagates up through the inline values handler. While the handler has a try/catch at line 155 that wraps the entire evaluation loop, the `formatValue` call is **inside** the inner try/catch (line 163), so a `TypeError` from `formatValue` would actually be caught by that inner catch and logged at debug level — meaning the error is silently swallowed rather than the variable being skipped gracefully. The user sees no inline value for that variable, but the error message in the debug log is misleading (it reports the expression evaluation error, not the type mismatch).

### Why This Matters

The bridge is a separate process. Bridge bugs, serialization failures, or Pike runtime edge cases can produce mismatched type/value pairs. The current code trusts the bridge's type annotation blindly. The codebase has a pattern of validating bridge responses before using them (see recent commits `81299d1`, `bf566d9` replacing `as-unknown-as` casts with typed access in `buildHoverContent`).

#### Change 1: Add runtime guards to the `array` branch (line 47)

**Before:**
```typescript
if (type === 'array') {
  const arr = value as unknown[];
  if (arr.length > 5) {
```

**After:**
```typescript
if (type === 'array' && Array.isArray(value)) {
  const arr = value as unknown[];
  if (arr.length > 5) {
```

Rationale: Move `Array.isArray(value)` into the condition. If the value is not an array, this branch is skipped entirely and execution falls through to the generic `String(value)` formatter at the bottom of the function. The cast `value as unknown[]` is now safe because TypeScript narrows after `Array.isArray`.

#### Change 2: Add runtime guard to the `mapping` branch (line 57)

**Before:**
```typescript
if (type === 'mapping') {
  const mapping = value as Record<string, unknown>;
  const entries = Object.entries(mapping);
```

**After:**
```typescript
if (type === 'mapping' && typeof value === 'object' && value !== null) {
  const mapping = value as Record<string, unknown>;
  const entries = Object.entries(mapping);
```

Rationale: `typeof value === 'object' && value !== null` ensures `value` is a non-null object before calling `Object.entries()`. This is the standard guard for objects that aren't arrays. Arrays also pass this check, but that's acceptable — `Object.entries` on an array returns index-keyed entries, and the display format `({k: v}, ...)` is close enough for a fallback case.

#### Change 3: Add runtime guard to the `multiset` branch (line 69)

**Before:**
```typescript
if (type === 'multiset') {
  const ms = value as unknown[];
  return `(${ms.join(', ')})`;
}
```

**After:**
```typescript
if (type === 'multiset' && Array.isArray(value)) {
  const ms = value as unknown[];
  return `(${ms.join(', ')})`;
}
```

Rationale: Same as the `array` branch — multiset values should be arrays at runtime.

#### Summary of changes

- 3 lines changed (adding `&&` guards to existing `if` conditions)
- 0 lines added
- 0 lines removed
- No new functions, types, or imports

#### Change 4: Add test cases for type/value mismatch

Add a new `describe('formatValue type mismatch handling')` block inside the existing test file. Tests exercise the handler end-to-end by having the bridge return mismatched type/value pairs.

**Test cases:**

1. **`type: 'array'` with `value: null`** — bridge returns `{ success: true, value: null, type: 'array' }`. Should not throw. Should fall through to `String(null)` = `"null"`.

2. **`type: 'array'` with `value: "not-an-array"`** — bridge returns `{ success: true, value: 'not-an-array', type: 'array' }`. Should fall through to `String("not-an-array")` = `"not-an-array"`.

3. **`type: 'mapping'` with `value: null`** — bridge returns `{ success: true, value: null, type: 'mapping' }`. Should fall through to `"null"`.

4. **`type: 'mapping'` with `value: 42`** — bridge returns `{ success: true, value: 42, type: 'mapping' }`. Should fall through to `"42"`.

5. **`type: 'multiset'` with `value: null`** — bridge returns `{ success: true, value: null, type: 'multiset' }`. Should fall through to `"null"`.

6. **`type: 'multiset'` with `value: "oops"`** — bridge returns `{ success: true, value: 'oops', type: 'multiset' }`. Should fall through to `"oops"`.

7. **`type: 'array'` with valid array** — regression: confirm existing behavior is preserved. Bridge returns `{ success: true, value: [1, 2, 3], type: 'array' }`. Should produce `"[1, 2, 3]"`.

Each test uses the existing `setupMock` helper and the `registerInlineValuesHandler` import pattern already established in the test file. The test asserts that the handler returns a result (does not throw) and that the `text` field contains the expected fallback string.

### LSP Latency

**Negligible.** The guards (`Array.isArray`, `typeof`, `=== null`) are O(1) microsecond-level operations. They execute only once per inline value result — the same frequency as the current casts. The inline values feature is already gated behind `globalSettings.inlineValues.enabled` and the visible-line range filter, so these guards never execute unless the feature is active.

### Memory

**Zero additional allocation.** No new objects, arrays, or closures are created. The guards are pure boolean checks on existing values.

### Cold Start

**No impact.** `formatValue` is defined inside `registerInlineValuesHandler` as a closure. The guards add no startup cost.

### Comparison to Alternatives

- **try/catch around each branch**: Would add overhead from exception creation (stack trace capture) on the failure path. Guard checks avoid this entirely.
- **Schema validation library**: Massive overkill for 3 type checks. Would add cold start cost and dependency weight.

### Unit tests (in existing test file)

| # | Scenario                              | Bridge returns                                    | Expected result         |
|---|---------------------------------------|---------------------------------------------------|--------------------------|
| 1 | array type, null value                | `{ value: null, type: 'array' }`                  | text contains `"null"`  |
| 2 | array type, string value              | `{ value: 'not-an-array', type: 'array' }`        | text contains value      |
| 3 | mapping type, null value              | `{ value: null, type: 'mapping' }`                | text contains `"null"`  |
| 4 | mapping type, number value            | `{ value: 42, type: 'mapping' }`                  | text contains `"42"`    |
| 5 | multiset type, null value             | `{ value: null, type: 'multiset' }`               | text contains `"null"`  |
| 6 | multiset type, string value           | `{ value: 'oops', type: 'multiset' }`             | text contains `"oops"`  |
| 7 | array type, valid array (regression)  | `{ value: [1, 2, 3], type: 'array' }`             | text contains `[1, 2, 3]`|

### Verification command

```bash
cd packages/pike-lsp-server && bun test src/tests/advanced/inline-values-provider.test.ts
```

## 5. Affected Files

| Path                                                          | Changes                                           |
|---------------------------------------------------------------|---------------------------------------------------|
| `packages/pike-lsp-server/src/features/advanced/inline-values.ts` | Add `&& Array.isArray(value)` guard on line 47 (`array` branch), add `&& typeof value === 'object' && value !== null` guard on line 57 (`mapping` branch), add `&& Array.isArray(value)` guard on line 69 (`multiset` branch). 3 lines modified, 0 lines added/removed. |
| `packages/pike-lsp-server/src/tests/advanced/inline-values-provider.test.ts` | Add `describe('formatValue type mismatch handling')` block with 7 test cases exercising mismatched type/value pairs from the bridge. ~80-100 lines added. |

TheSmuks:
Closes #2023

## Summary

1. Remove the describe block from inside the `it` 2. Insert it after the `it` closes (after line 973)I need to remove the describe block from its current position (lines 649-973) and place it after line 973 (the `it`'s closing).Missing a closing `});` for the outer describe. Let me check:The `it` at line 598 has its closing `});` at line 649, but this prematurely closes the outer `describe` block. The `it` should close with `});` then the new `describe` should be inside the outer describe. The problem is that the `it`'s `});` at 649 is actually closing the outer `describe` (started at line 82). I need to add an extra `});` for the `it`.Wait, now line 652 is `});` which closes the outer describe. Let me check:Now line 649 closes the `it`, line 652 closes the outer `describe`, and the new `describe` at 655 is outside the outer `describe`. I need to swap the order — the new describe should go before the outer describe closes.Let me re-read the relevant sections to get accurate anchors:I need to: 1. Remove lines 652-654 (`});`, empty, empty)

## Linked Issue

Closes #2023

## Root Cause

The formatValue() function in inline-values.ts casts the value parameter to unknown[] (line 47, for type 'array') and to Record<string,unknown> (line 69, for type 'multiset') based solely on the type string from the bridge. There is no runtime validation (e.g., Array.isArray) before the cast. If the bridge returns a type annotation of 'array' but the actual value is not an array (e.g., null or a string due to a bridge bug or edge case), the function will throw an unhandled TypeError when accessing .length or .slice().

## Changes

- .agent-knowledge/reference/KB-2023.md: (see commit diffs)
- packages/pike-lsp-server/src/features/advanced/inline-values.ts: (see commit diffs)
- packages/pike-lsp-server/src/tests/advanced/inline-values-provider.test.ts: (see commit diffs)

## Knowledge Base

- KB ID: KB-AUTO-2023

## Verification

## Checklist

## Notes for Reviewer

### DIFF
--- a/.agent-knowledge/reference/KB-2023.md
+++ b/.agent-knowledge/reference/KB-2023.md
@@ -1,0 +1,18 @@
+---
+id: KB-AUTO-2023
+domain: REFACTOR
+date: 2026-04-15
+authors: [dga-coder]
+summary: Added runtime type guards before unsafe casts in formatValue() to prevent TypeErrors from mismatched bridge type annotations
+---
+
+# KB-2023: Add runtime type guards before unsafe casts in formatValue()
+
+## Summary
+
+The `formatValue()` function in `inline-values.ts` cast `value: unknown` to `unknown[]` or `Record<string, unknown>` based solely on the bridge's `type` string, without validating the actual value shape. If the bridge returned a mismatched type/value pair (e.g., `type: 'array'` but `value: null`), the function would throw an unhandled TypeError. Added `Array.isArray()` guards for array/multiset branches and `typeof value === 'object' && value !== null` for the mapping branch, causing mismatches to fall through to the generic `String(value)` formatter.
+
+## Key Files Changed
+
+- packages/pike-lsp-server/src/features/advanced/inline-values.ts: Added `&& Array.isArray(value)` guard on `array` branch, `&& typeof value === 'object' && value !== null` on `mapping` branch, `&& Array.isArray(value)` on `multiset` branch.
+- packages/pike-lsp-server/src/tests/advanced/inline-values-provider.test.ts: Added 7 test cases for type/value mismatch handling (array+null, array+string, mapping+null, mapping+number, multiset+null, multiset+string, valid array regression).
--- a/packages/pike-lsp-server/src/features/advanced/inline-values.ts
+++ b/packages/pike-lsp-server/src/features/advanced/inline-values.ts
@@ -43,7 +43,7 @@ export function registerInlineValuesHandler(
     if (type === 'int' || type === 'float') {
       return String(value);
     }
-    if (type === 'array') {
+    if (type === 'array' && Array.isArray(value)) {
       const arr = value as unknown[];
       if (arr.length > 5) {
         return `[${arr
@@ -53,7 +53,7 @@ export function registerInlineValuesHandler(
       }
       return `[${arr.map(v => formatSimpleValue(v)).join(', ')}]`;
     }
-    if (type === 'mapping') {
+    if (type === 'mapping' && typeof value === 'object' && value !== null) {
       const mapping = value as Record<string, unknown>;
       const entries = Object.entries(mapping);
       if (entries.length > 3) {
@@ -65,7 +65,7 @@ export function registerInlineValuesHandler(
       }
       return `(${entries.map(([k, v]) => `${k}: ${formatSimpleValue(v)}`).join(', ')})`;
     }
-    if (type === 'multiset') {
+    if (type === 'multiset' && Array.isArray(value)) {
       const ms = value as unknown[];
       return `(${ms.join(', ')})`;
     }
--- a/packages/pike-lsp-server/src/tests/advanced/inline-values-provider.test.ts
+++ b/packages/pike-lsp-server/src/tests/advanced/inline-values-provider.test.ts
@@ -647,4 +647,352 @@ describe('Inline Values Provider', () => {
     assert.strictEqual(result.length, 1);
     assert.ok(result[0]!.text.includes('42'), `Expected "42" in "${result[0]!.text}"`);
   });
+
+  describe('formatValue type mismatch handling', () => {
+    it('should fall through to String() when array type has null value', async () => {
+      const { registerInlineValuesHandler } =
+        await import('../../features/advanced/inline-values.js');
+
+      const code = 'int x = 0;';
+      const services: MockServices = {
+        globalSettings: { inlineValues: { enabled: true } },
+        documentCache: {
+          get: () => ({
+            symbols: [
+              {
+                kind: 'variable',
+                name: 'x',
+                range: { start: { line: 0, character: 4 }, end: { line: 0, character: 11 } },
+                selectionRange: {
+                  start: { line: 0, character: 4 },
+                  end: { line: 0, character: 5 },
+                },
+              },
+            ],
+            diagnostics: [],
+          }),
+        },
+        bridge: {
+          bridge: {
+            async evaluateConstant() {
+              return { success: true, value: null, type: 'array' };
+            },
+          },
+        },
+      };
+
+      const { connection, documents, getHandler } = setupMock(services, code);
+      registerInlineValuesHandler(connection, services, documents);
+
+      const params: InlineValueParams = {
+        textDocument: { uri: 'file:///test.pike' },
+        range: { start: { line: 0, character: 0 }, end: { line: 10, character: 0 } },
+        context: {
+          frameId: 0,
+          stoppedLocation: { start: { line: 0, character: 0 }, end: { line: 10, character: 0 } },
+        },
+      };
+
+      const result = await getHandler()!(params);
+      assert.ok(result, 'Should return inline values');
+      assert.ok(result[0]!.text.includes('NULL'), `Expected "NULL" in "${result[0]!.text}"`);
+    });
+
+    it('should fall through to String() when array type has string value', async () => {
+      const { registerInlineValuesHandler } =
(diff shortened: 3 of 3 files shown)


TheSmuks:
**DGA: vote-approved**

Merge vote: 3/3 approved.
- **reviewer**: APPROVE
- **pike-voter**: APPROVE

PR #2034, opened 2026-04-16
fix(#2033): Fix clipEditToRange using original text line length for newT

### LINKED ISSUE #2033: fix: Fix clipEditToRange using original text line length for newText clipping
## Description

Fix clipEditToRange using original text line length for newText clipping

## Required Labels

Priority: P2 | Type: bug

## Problem

In clipEditToRange(), when an edit extends beyond endLine, clippedEndCol is set to endLineLen which is the length of the original text's line at endLine. This value is then used to slice the last visible newText line: `visibleLines[...].slice(0, clippedEndCol)`. If the formatter's newText has longer lines than the original text (e.g., operator spacing adds spaces), those extra characters are silently dropped. This causes the clipped edit to produce incorrect output — the resulting text at endLine would be truncated.

## Expected Behavior

When clipping an edit that extends beyond the format range, the last visible newText line should be kept in full (or clipped only at the explicit edit end character), not truncated to the original text's line length.

## Suggested Approach

Change the clippedEndCol logic: when edit.range.end.line > endLine, set clippedEndCol to visibleLines[visibleLines.length - 1].length (the full newText line) instead of endLineLen. The purpose of clipping is to drop lines outside the range, not to truncate characters within the last kept line. Update the slice call accordingly.

## Affected Files

packages/pike-lsp-server/src/services/formatting-service.ts:165

## Acceptance

- [ ] Code change implemented
- [ ] Tests added/updated

**Priority:** P2 | **Type:** bug | Auto-discovered by DGA Architect

## Implementation Spec

# Spec: Fix clipEditToRange truncating last visible newText line

### Root Cause

**File**: `packages/pike-lsp-server/src/services/formatting-service.ts`
**Function**: `clipEditToRange` (line 148)
**Buggy line**: Line 166

```typescript
const clippedEndCol = edit.range.end.line > endLine ? endLineLen : edit.range.end.character;
```

When an edit extends beyond `endLine` (i.e., `edit.range.end.line > endLine`), `clippedEndCol` is set to `endLineLen`, which is the length of the **original document's** line at `endLine` (computed at line 161: `const endLineLen = lines[endLine]?.length ?? 0;`).

This `clippedEndCol` is then used to truncate the last visible newText line at three points:

- **Line 175** (single visible line): `visibleLines[0]!.slice(clippedStartCol, clippedEndCol)`
- **Line 180** (multi-line, last line): `visibleLines[visibleLines.length - 1]!.slice(0, clippedEndCol)`

The problem: `endLineLen` is a property of the **original text**, not the **formatted newText**. When the formatter produces newText with longer lines than the original (e.g., `applyOperatorSpacing` inserts spaces around operators like `a+b` → `a + b`), the last visible newText line can be longer than `endLineLen`. Slicing to `endLineLen` silently drops those extra characters.

### Concrete Example

Original text:
```
line 0: class Foo {
line 1:   int a=1+2;
line 2: }
```

Formatter with `spaceAroundOperators: true` produces newText `  int a = 1 + 2;` (length 17) for lines 0-2.

If `formatRange` is called with `startLine=0, endLine=1`:
- `endLineLen = lines[1].length = 13` (length of `  int a=1+2;`)
- The edit spans lines 0-2, so it gets clipped to end at line 1
- `clippedEndCol = 13`
- The last visible newText line `  int a = 1 + 2;` (17 chars) gets sliced to `  int a = 1 + 2` (13 chars) — **dropping the trailing semicolon**

### Why This Is Wrong

The purpose of clipping is to **remove whole lines outside the requested range**, not to truncate characters within the last kept line. When an edit extends beyond `endLine`, the correct behavior is:
1. Drop newText lines that correspond to document lines beyond `endLine` (already done via `skipCount`)
2. Keep the last visible newText line **in full** — because it corresponds to `endLine`, and the formatter's replacement for that line should not be truncated to the original line's length

The `endLineLen` value is used correctly for `endPos` (the range boundary in the document), but incorrectly for `clippedEndCol` (the newText boundary). These serve different purposes: `endPos` says where in the document to stop replacing, while `clippedEndCol` says how much of the newText to keep.

### File 1: `packages/pike-lsp-server/src/services/formatting-service.ts`

**Change 1: Fix `clippedEndCol` computation (line 166)**

Current:
```typescript
const clippedEndCol = edit.range.end.line > endLine ? endLineLen : edit.range.end.character;
```

Replace with:
```typescript
const newTextLines = edit.newText.split('\n');
const clippedEndCol = edit.range.end.line > endLine
  ? (newTextLines[newTextLines.length - 1]?.length ?? 0)
  : edit.range.end.character;
```

This moves the `newText.split('\n')` call up (it was on line 167, now used before) and changes the fallback from `endLineLen` (original text line length) to the last newText line's actual length. When the edit extends beyond `endLine`, we keep the full last visible newText line rather than truncating to the original text's line length.

**Change 2: Remove the now-duplicate `newTextLines` declaration (line 167)**

The `const newTextLines = edit.newText.split('\n');` on line 167 becomes redundant after Change 1 moves it up. Remove it.

**Net effect on the function body** (lines 165-167 before → after):

Before:
```typescript
  const clippedStartCol = edit.range.start.line < startLine ? 0 : edit.range.start.character;
  const clippedEndCol = edit.range.end.line > endLine ? endLineLen : edit.range.end.character;
  const newTextLines = edit.newText.split('\n');
```

After:
```typescript
  const clippedStartCol = edit.range.start.line < startLine ? 0 : edit.range.start.character;
  const newTextLines = edit.newText.split('\n');
  const clippedEndCol = edit.range.end.line > endLine
    ? (newTextLines[newTextLines.length - 1]?.length ?? 0)
    : edit.range.end.character;
```

**No other changes needed.** The `endPos` computation (line 161-162) correctly uses `endLineLen` for the document range boundary — that stays as-is. The `clippedStartCol` computation (line 165) is correct — that stays as-is. All downstream usage of `clippedEndCol` in the slicing logic (lines 175, 180) now receives the correct value.

### File 2: `packages/pike-lsp-server/src/tests/services/formatting-service.test.ts`

**Add a new test case** inside the `clipEditToRange (via formatRange)` describe block that specifically verifies the bug scenario — an edit whose newText has a longer last line than the original text's endLine.

```typescript
it('does not truncate last visible newText line when edit extends beyond endLine', () => {
  const service = new FormattingService();
  const text = [
    'class Foo {',           // line 0
    '  int a=1+2;',           // line 1 — operator spacing will widen this
    '  int b=3+4;',           // line 2
    '}',                      // line 3
  ].join('\n');

  // Request range [0, 1]. Operator spacing edits span the full document.
  // The edit replacing lines 0-3 has newText where line 1 content is wider
  // than the original (e.g., '  int a = 1 + 2;' vs '  int a=1+2;').
  // The clipped edit's newText must not be truncated.
  const rangeEdits = service.formatRange(text, 0, 1, {});
  const fullEdits = service.formatDocument(text, {});

  // Simulate applying edits: the result for lines 0-1 must match what
  // full-document formatting produces for those same lines.
  const lines = text.split('\n');
  let rangeResult = lines.slice(0, 2).join('\n');
  // Apply range edits in reverse to preserve positions
  const sortedRangeEdits = [...rangeEdits].sort(
    (a, b) => b.range.start.line - a.range.start.line || b.range.start.character - a.range.start.character
  );
  for (const edit of sortedRangeEdits) {
    const resultLines = rangeResult.split('\n');
    if (edit.range.start.line === edit.range.end.line) {
      const line = resultLines[edit.range.start.line] ?? '';
      resultLines[edit.range.start.line] =
        line.slice(0, edit.range.start.character) +
        edit.newText +
        line.slice(edit.range.end.character);
      rangeResult = resultLines.join('\n');
    }
  }

  // Full-format the first two lines independently for comparison
  let fullResult = lines.slice(0, 2).join('\n');
  const fullEditsInRange = fullEdits.filter(
    e => e.range.start.line <= 1 && e.range.end.line >= 0
  );
  const sortedFullEdits = [...fullEditsInRange].sort(
    (a, b) => b.range.start.line - a.range.start.line || b.range.start.character - a.range.start.character
  );
  for (const edit of sortedFullEdits) {
    const resultLines = fullResult.split('\n');
    if (edit.range.start.line <= 1 && edit.range.end.line <= 1) {
      const line = resultLines[edit.range.start.line] ?? '';
      const endChar = edit.range.end.line === edit.range.start.line
        ? Math.min(edit.range.end.character, line.length)
        : line.length;
      resultLines[edit.range.start.line] =
        line.slice(0, edit.range.start.character) +
        edit.newText +
        line.slice(endChar);
      fullResult = resultLines.join('\n');
    }
  }

  assert.strictEqual(
    rangeResult.split('\n')[1],
    fullResult.split('\n')[1],
    'Range-formatted line 1 must match fully-formatted line 1 (not truncated)'
  );
});
```

## 3. Performance Considerations

**LSP Latency**: Negligible. The change replaces `endLineLen` (an already-computed integer) with `newTextLines[newTextLines.length - 1]?.length` (an array index + property access). The `newText.split('\n')` call is moved up by one line but still executes exactly once per call. No additional allocations.

**Memory**: Zero change. The same `newTextLines` array is created; we just read from it slightly earlier.

**Cold Start**: Zero impact. No new imports, no new modules, no initialization logic.

### New test: `does not truncate last visible newText line when edit extends beyond endLine`
- **Location**: `packages/pike-lsp-server/src/tests/formatting-service.test.ts`, inside `clipEditToRange (via formatRange)` describe
- **Setup**: Document with unspaced operators (`a=1+2;`), format range `[0, 1]` with default profile (`spaceAroundOperators: true`)
- **Assertion**: The formatted line 1 from range formatting matches formatted line 1 from full-document formatting — specifically, the trailing semicolon is preserved
- **Why this catches the bug**: With the bug, `clippedEndCol = endLineLen = 13` truncates `  int a = 1 + 2;` (17 chars) to 13 chars, dropping the semicolon. After the fix, `clippedEndCol` equals the full newText line length (17), preserving the semicolon.

### Regression: existing tests must still pass
All 8 existing `clipEditToRange` tests plus all other formatting tests continue to pass. The fix is strictly more permissive (keeps more of the newText), so tests that assert correct clipping boundaries remain valid.

### Verification command
```bash
cd packages/pike-lsp-server && bun test src/tests/services/formatting-service.test.ts
```

## 5. Affected Files

| File | Change |
|------|--------|
| `packages/pike-lsp-server/src/services/formatting-service.ts` | Move `newTextLines` declaration before `clippedEndCol`; change `clippedEndCol` to use last newText line length instead of `endLineLen` when edit extends beyond `endLine` |
| `packages/pike-lsp-server/src/tests/services/formatting-service.test.ts` | Add 1 new test case verifying no truncation when newText is longer than original text at endLine |

TheSmuks:
Closes #2033

## Summary

Now I have a clear picture of the bug and the fix. Let me implement it. **Edit 1: Fix `clippedEndCol` in formatting-service.ts****Edit 2: Add the test case**Now run typecheck and tests:All 23 tests pass (22 existing + 1 new). Now write the KB entry:Done. The fix changes `clippedEndCol` from using `endLineLen` (original text) to `newTextLines[newTextLines.length - 1]?.length` (the actual newText line) when an edit extends beyond the format range. This prevents silent truncation of formatted output that is wider than the original text (e.g., operator spacing).

## Linked Issue

Closes #2033

## Root Cause

In clipEditToRange(), when an edit extends beyond endLine, clippedEndCol is set to endLineLen which is the length of the original text's line at endLine. This value is then used to slice the last visible newText line: `visibleLines[...].slice(0, clippedEndCol)`. If the formatter's newText has longer lines than the original text (e.g., operator spacing adds spaces), those extra characters are silently dropped. This causes the clipped edit to produce incorrect output — the resulting text at endLine would be truncated.

## Changes

- .agent-knowledge/reference/KB-2033.md: (see commit diffs)
- packages/pike-lsp-server/src/services/formatting-service.ts: (see commit diffs)
- packages/pike-lsp-server/src/tests/services/formatting-service.test.ts: (see commit diffs)

## Knowledge Base

- KB ID: KB-AUTO-2033

## Verification

## Checklist

## Notes for Reviewer

### DIFF
--- a/.agent-knowledge/reference/KB-2033.md
+++ b/.agent-knowledge/reference/KB-2033.md
@@ -1,0 +1,18 @@
+---
+id: KB-AUTO-2033
+domain: REFACTOR
+date: 2026-04-16
+authors: [dga-coder]
+summary: Fixed clipEditToRange using original text line length instead of newText line length when clipping edits that extend beyond the format range
+---
+
+# KB-2033: Fix clipEditToRange truncating newText at original text line length
+
+## Summary
+
+`clipEditToRange()` in `formatting-service.ts` used `endLineLen` (the original document's line length) to slice the last visible `newText` line when an edit extended beyond `endLine`. When the formatter's newText had longer lines than the original (e.g., operator spacing `a+b` → `a + b`), the extra characters were silently dropped. Fixed by using the last `newText` line's actual length instead of `endLineLen`.
+
+## Key Files Changed
+
+- `packages/pike-lsp-server/src/services/formatting-service.ts`: Moved `newTextLines` declaration before `clippedEndCol`; changed `clippedEndCol` to use `newTextLines[newTextLines.length - 1]?.length` instead of `endLineLen` when `edit.range.end.line > endLine`
+- `packages/pike-lsp-server/src/tests/services/formatting-service.test.ts`: Added test verifying range-formatted output matches full-document formatting when newText lines are longer than original
--- a/packages/pike-lsp-server/src/services/formatting-service.ts
+++ b/packages/pike-lsp-server/src/services/formatting-service.ts
@@ -164,8 +164,11 @@ function clipEditToRange(
 
   // Compute the portion of newText that falls within the clipped range
   const clippedStartCol = edit.range.start.line < startLine ? 0 : edit.range.start.character;
-  const clippedEndCol = edit.range.end.line > endLine ? endLineLen : edit.range.end.character;
   const newTextLines = edit.newText.split('\n');
+  const clippedEndCol =
+    edit.range.end.line > endLine
+      ? (newTextLines[newTextLines.length - 1]?.length ?? 0)
+      : edit.range.end.character;
   let clippedNewText: string;
 
   if (edit.range.start.line === edit.range.end.line) {
--- a/packages/pike-lsp-server/src/tests/services/formatting-service.test.ts
+++ b/packages/pike-lsp-server/src/tests/services/formatting-service.test.ts
@@ -249,6 +249,75 @@ describe('FormattingService', () => {
         `Multi-line edit spans ${editLineSpan} lines but newText has ${fullNewTextLines.length} lines`
       );
     });
+
+    it('does not truncate last visible newText line when edit extends beyond endLine', () => {
+      const service = new FormattingService();
+      const text = [
+        'class Foo {', // line 0
+        '  int a=1+2;', // line 1 — operator spacing will widen this
+        '  int b=3+4;', // line 2
+        '}', // line 3
+      ].join('\n');
+
+      // Request range [0, 1]. Operator spacing edits span the full document.
+      // The edit replacing lines 0-3 has newText where line 1 content is wider
+      // than the original (e.g., '  int a = 1 + 2;' vs '  int a=1+2;').
+      // The clipped edit's newText must not be truncated.
+      const rangeEdits = service.formatRange(text, 0, 1, {});
+      const fullEdits = service.formatDocument(text, {});
+
+      // Simulate applying edits: the result for lines 0-1 must match what
+      // full-document formatting produces for those same lines.
+      const lines = text.split('\n');
+      let rangeResult = lines.slice(0, 2).join('\n');
+      // Apply range edits in reverse to preserve positions
+      const sortedRangeEdits = [...rangeEdits].sort(
+        (a, b) =>
+          b.range.start.line - a.range.start.line ||
+          b.range.start.character - a.range.start.character
+      );
+      for (const edit of sortedRangeEdits) {
+        const resultLines = rangeResult.split('\n');
+        if (edit.range.start.line === edit.range.end.line) {
+          const line = resultLines[edit.range.start.line] ?? '';
+          resultLines[edit.range.start.line] =
+            line.slice(0, edit.range.start.character) +
+            edit.newText +
+            line.slice(edit.range.end.character);
+          rangeResult = resultLines.join('\n');
+        }
+      }
+
+      // Full-format the first two lines independently for comparison
+      let fullResult = lines.slice(0, 2).join('\n');
+      const fullEditsInRange = fullEdits.filter(
+        e => e.range.start.line <= 1 && e.range.end.line >= 0
+      );
+      const sortedFullEdits = [...fullEditsInRange].sort(
+        (a, b) =>
+          b.range.start.line - a.range.start.line ||
+          b.range.start.character - a.range.start.character
+      );
+      for (const edit of sortedFullEdits) {
+        const resultLines = fullResult.split('\n');
+        if (edit.range.start.line <= 1 && edit.range.end.line <= 1) {
+          const line = resultLines[edit.range.start.line] ?? '';
+          const endChar =
+            edit.range.end.line === edit.range.start.line
+              ? Math.min(edit.range.end.character, line.length)
+              : line.length;
+          resultLines[edit.range.start.line] =
+            line.slice(0, edit.range.start.character) + edit.newText + line.slice(endChar);
+          fullResult = resultLines.join('\n');
+        }
+      }
+
+      assert.strictEqual(
+        rangeResult.split('\n')[1],
+        fullResult.split('\n')[1],
+        'Range-formatted line 1 must match fully-formatted line 1 (not truncated)'
+      );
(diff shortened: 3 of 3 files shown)


TheSmuks:
**DGA: vote-approved**

Merge vote: 3/3 approved.
- **reviewer**: APPROVE
- **pike-voter**: APPROVE

PR #2214, opened 2026-04-20
fix(#2207): Fix or remove getInheritedMembers — it finds same-name symbo

### LINKED ISSUE #2207: fix: Fix or remove getInheritedMembers — it finds same-name symbols not inherited ones
## Description

Fix or remove getInheritedMembers — it finds same-name symbols not inherited ones

## Required Labels

Priority: P0-broken | Type: bug

## Problem

getInheritedMembers claims to trace through inherit chain but actually just returns all symbols with the same name as the input symbol. It has no inherit-chain logic at all. Callers using this for rename scoping will rename symbols with coincidentally identical names in unrelated scopes.

## Expected Behavior

The code handles this case correctly.

## Suggested Approach

Either (a) remove the function and its callers if inherited-member rename is not yet supported, or (b) implement actual inherit-chain resolution using the workspace index inherit edges. Verify callers in rename.ts to determine which path.

## Affected Files

packages/pike-lsp-server/src/features/editing/rename.ts:391-409

## Acceptance

- [ ] Code change implemented
- [ ] Tests added/updated

## Environment

- Pike binary: 8.0.1116
- Bun version: 1.2.x
- $PIKE_SRC set: YES
- $ROXEN_SRC set: YES

## Implementation Spec

# Implementation Spec: Remove Broken getInheritedMembers from rename.ts

### The broken function
**File**: `packages/pike-lsp-server/src/features/editing/rename.ts`
**Lines**: 391–409

```typescript
function getInheritedMembers(
  symbol: { name: string; kind?: string; position?: { line: number } },
  allSymbols: Array<{ name: string; kind?: string; position?: { line: number } }>
): Array<{ name: string; kind?: string; position?: { line: number } }> {
  const inherited: Array<{ name: string; kind?: string; position?: { line: number } }> = [];
  for (const s of allSymbols) {
    if (s.name === symbol.name && s !== symbol) {
      inherited.push(s);
    }
  }
  return inherited;
}
```

**What it claims to do** (doc comment, line 392): "Traces through inherit chain to find members with the same name."

**What it actually does**: Returns every symbol in the `allSymbols` array that has the same `.name` as the input symbol, excluding only the input symbol itself. There is zero inherit-chain logic. No `inherits` edges are consulted, no class hierarchy is walked, no scope containment is checked.

**Impact on callers**: The only call site (line 244) passes `cached?.symbols ?? []` as `allSymbols`. This means any symbol in the document with a coincidentally identical name — in a completely unrelated scope, class, or namespace — will be reported as an "inherited member." If this result were ever used for rename scoping (it currently only produces a debug log, but the intent is clearly to grow into real rename logic), it would cause false-positive renames across unrelated scopes.

### The only call site
**Lines**: 244–250
```typescript
if (matchingSymbol) {
  const inheritedMembers = getInheritedMembers(matchingSymbol, cached?.symbols ?? []);
  if (inheritedMembers.length > 0) {
    log.debug('Rename: symbol has inherited members', {
      count: inheritedMembers.length,
    });
  }
}
```

**Current behavioral impact**: This block only emits a debug log. It does not affect rename edit generation. The rename logic below it (lines 252+) uses `symbolPositions`, token-based search, or text search — none of which depend on `getInheritedMembers`. So the broken function produces misleading log output but does not corrupt rename edits today.

**Why it must still be removed**: (1) The debug log is semantically wrong — it claims inherited members exist when it's really finding name-collisions across unrelated scopes. This misleads developers debugging rename issues. (2) The function's existence and doc comment signal that inherited-member rename is supported, which is false. (3) Leaving dead code with incorrect semantics invites future developers to wire it into actual rename logic, introducing the cross-scope rename bug.

### Misleading file header
**Line 13**: ` * - Inherited member renaming support`

This feature does not exist. The header should not claim it.

### Approach: Remove the function and its callers (Option A from the issue)

The issue offers two approaches: (a) remove the function and its callers, or (b) implement actual inherit-chain resolution. The correct choice is (a), because:

1. **The result is unused.** The function only produces a debug log. No rename edits depend on it.
2. **Proper implementation exists elsewhere.** `TypeDatabase.getInheritedMembers()` at `type-database.ts:179` correctly walks the `inheritanceGraph` and recursively resolves ancestor members. If inherited-member rename is ever implemented, it should use `services.typeDatabase`, not a local function.
3. **The document cache has the inherit edges.** `DocumentCacheEntry.inherits` (core/types.ts:204) stores `InheritanceInfo[]` from introspection. A correct implementation would use these edges, not a flat name-match.
4. **The function's signature is wrong for real use.** It accepts bare `{ name, kind?, position? }` objects, losing all type information needed for proper scope resolution.

#### File 1: `packages/pike-lsp-server/src/features/editing/rename.ts`

**Change A** — Remove file header claim (line 13)
- Delete the line: ` * - Inherited member renaming support`

**Change B** — Remove the call site block (lines 243–250)
- Delete the entire `if (matchingSymbol) { ... }` block that calls `getInheritedMembers` and logs the count. The `matchingSymbol` variable is still used later (line 267, 285) so only this 8-line block is removed, not the variable declaration.

**Change C** — Remove the function definition (lines 391–409)
- Delete `getInheritedMembers` and its JSDoc comment entirely.

#### File 2: `packages/pike-lsp-server/src/scenarios/rename-cross-file.test.ts`

**Change D** — Remove the "Inherited member rename" describe block (lines 129–167)
- This contains two tests (`'should track inherited members through class hierarchy'` and `'should rename inherited members across the hierarchy'`) that test the intended behavior of the broken function using inline mock data. They don't test actual rename handler code — they construct plain objects and run Array.filter on them. They validate the broken name-matching logic, not real inherit-chain resolution.
- The rest of the test file (collision detection, prepare rename validation, workspace symbol resolution, deterministic rename, result format) is unaffected and remains valid.

### LSP latency
**Impact: None.** The removed code only produced a debug log. Removing it eliminates one O(n) iteration over `cached.symbols` on every rename request where `matchingSymbol` is found, but this iteration is trivially cheap (typically <100 symbols per document) and was never on any critical path.

### Memory
**Impact: None.** No data structures were allocated persistently. The `inherited` array was local and immediately discarded after logging.

### Cold start
**Impact: None.** `getInheritedMembers` is a module-level function definition. Removing it reduces the module's parsed size by ~20 lines, which is negligible.

### Test Case 1: Rename still works after function removal

**GIVEN** a Pike file containing a function `sharedHelper` that is also referenced in another file
**WHEN** the user invokes rename on `sharedHelper` with new name `betterHelper`
**THEN** the rename completes successfully with edits in both files, and no debug log about "inherited members" is emitted

*Verification*: Run the existing rename scenario tests in `rename-cross-file.test.ts`. All tests outside the removed "Inherited member rename" block must pass.

### Test Case 2: No false "inherited members" log

**GIVEN** a Pike file with two symbols named `x` in different scopes (e.g., global `x` and local `x` inside a function)
**WHEN** the user renames one of them
**THEN** the debug log does NOT contain a spurious "symbol has inherited members" message

*Verification*: Manual check or add a unit test that verifies the rename handler's log output does not mention "inherited members" for this case.

### Test Case 3: Removed test file section does not break suite

**GIVEN** the "Inherited member rename" describe block is removed from `rename-cross-file.test.ts`
**WHEN** the test suite runs
**THEN** all remaining tests pass (collision detection, prepare rename, workspace resolution, deterministic rename, result format)

*Verification*: `bun test packages/pike-lsp-server/src/scenarios/rename-cross-file.test.ts`

### Test Case 4: No remaining references to removed function

**GIVEN** `getInheritedMembers` is deleted from `rename.ts`
**WHEN** searching for `getInheritedMembers` across `packages/pike-lsp-server/src/features/editing/`
**THEN** zero results are found

*Verification*: `grep -r getInheritedMembers packages/pike-lsp-server/src/features/editing/` returns nothing.

## 5. Affected Files

| File | Lines Changed | Change Description |
|------|--------------|-------------------|
| `packages/pike-lsp-server/src/features/editing/rename.ts` | Delete L13, L243–250, L391–409 | Remove broken function, its call site, and misleading header |
| `packages/pike-lsp-server/src/scenarios/rename-cross-file.test.ts` | Delete L129–167 | Remove tests for the broken function's intended behavior |

## 6. Future Work (Not In Scope)

When inherited-member rename is genuinely needed:
1. Use `services.typeDatabase.getInheritedMembers(className)` which has correct inherit-chain resolution via the `inheritanceGraph` Map.
2. Use `cached?.inherits` (`InheritanceInfo[]`) to scope resolution to the current document's actual inherit edges.
3. Wire the results into the rename edit generation (currently lines 267+), not just a debug log.
4. Write integration tests that exercise actual Pike class hierarchies via the bridge, not inline mock objects.

TheSmuks:
Closes #2207

## Summary

Implements fix: Fix or remove getInheritedMembers — it finds same-name symbols not inherited ones

## Linked Issue

Closes #2207

## Root Cause

getInheritedMembers claims to trace through inherit chain but actually just returns all symbols with the same name as the input symbol. It has no inherit-chain logic at all. Callers using this for rename scoping will rename symbols with coincidentally identical names in unrelated scopes.

## Changes

- .agent-knowledge/reference/KB-2203-review.md: (see commit diffs)
- .agent-knowledge/reference/KB-2203.md: (see commit diffs)
- .agent-knowledge/reference/KB-2204.md: (see commit diffs)
- .agent-knowledge/reference/KB-2205.md: (see commit diffs)
- .agent-knowledge/reference/KB-2207.md: (see commit diffs)
- packages/pike-lsp-server/src/features/diagnostics/lifecycle.ts: (see commit diffs)
- packages/pike-lsp-server/src/features/editing/rename.ts: (see commit diffs)
- packages/pike-lsp-server/src/scenarios/rename-cross-file.test.ts: (see commit diffs)
- packages/pike-lsp-server/src/services/pike-introspection.ts: (see commit diffs)
- test/lib/punit: (see commit diffs)

## Knowledge Base

- KB ID: KB-AUTO-2207

## Verification

## Checklist

## Notes for Reviewer


TheSmuks:
**DGA: vote-approved**

Merge vote: 2/2 approved.
- APPROVE
- APPROVE

